PR #1, opened 2013-03-12
Handle login failures

gmokki:
The code was missing notify() when server returned login failure. The returned error was also ignored resulting in a non-working connection returned to user application.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
 		<dependency>
 			<groupId>io.netty</groupId>
 			<artifactId>netty</artifactId>
-			<version>3.6.2.Final</version>
+			<version>3.6.3.Final</version>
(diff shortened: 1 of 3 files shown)


kdubb:
Thanks for the commits.  Nice to see somebody jump in with both feet

I was trying to keep java.sql.\* type isolated to the jdbc package.  I am not certain that is the way to go throughout but in this case init is wrapped by PSQLDriver so throwing an IOException would be caught and appropriately translated to an SQLException.  Can you switch it use IOException instead?  I'd like to give you the honors on the first pull request merge!

kdubb:
I did the change myself as I saw the need for a new NoticeException that can easily shuttle PostgreSQL's notices around and be easily translated into good SQLExceptions later on.

PR #2, opened 2013-03-21
Support unicode table and field names

gmokki:
Example DDL that works after the patch:

``` sql
create table a€ (val€ int);
```


kdubb:
It occurs to me that US_ASCII is pretty much the wrong charset to use all the time.  We need to find out if C strings in the FE/BE protocol are always in UTF8 (since your changes are working) or if they should be written in the charset provided by "client_encoding".  I'll look into this ASAP and act accordingly.

kdubb:
After reading mailing list archives, it seems we should be respecting client encoding when reading/writing C strings.  I checked in a fix to the main development branch that stops assuming a charset for them.  This should fix your problem.

PR #5, opened 2013-03-30
Remember described result fields.

gmokki:
The main postgres jdbc driver does the same caching.


gmokki:
The commit 207b9963af9b4c820a3d9df917d6bee9bf2f421a seems to have been the correct fix. This can be ignored.

PR #6, opened 2013-03-31
Combine protocol messages to one ChannelBuffer and only write to socket once.

gmokki:
This patch is for the prepared statement execute.

In a my benchmark it speeds up a simple query with two result rows in a loop to localhost postgresql server by upto 15x by avoiding multiple system calls per round-trip.


kdubb:
Awesome work!  The is good stuff!

After integrating this pull request I saw the same large increase in speed that you mentioned. Seeing that I saw that I could do quite a bit more message combining and rewrote the protocol and commands to use this method explicitly.  It copies your idea but is more explicit about what is writing, sending, allocating buffers and renames the methods to match what they actually do.  Now all messages in a command are sent in a single channel write. 

Thanks!

PR #16, opened 2013-10-26
Initial checkstyle integration

jesperpedersen:
Mainly whitespace fixes, and import reordering


kdubb:
Is there a way to allow the "if (x == false) ..." format instead of requiring it to be "if (!x)"?

There are times when, to me, it seems clearer to use the "== false" format. For example, "if ( !(x instanceof Type) )" seems less obvious, and harder to type, than the "== false" version.  I don't see a need revert those cases of this in the pull request but I'd like to be able to use it in the future.

jesperpedersen:
Branch updated

PR #19, opened 2013-10-29
Remove finalizers

jesperpedersen:
Rely on the applications closes themselves, or the managed environment such as an application server.

### DIFF
--- a/src/main/java/com/impossibl/postgres/jdbc/PGConnection.java
+++ b/src/main/java/com/impossibl/postgres/jdbc/PGConnection.java
@@ -137,10 +137,6 @@ class PGConnection extends BasicContext implements Connection {
     }
   }
 
-  protected void finalize() throws SQLException {
-    close();
-  }
-
   /**
    * Add warning to end of warning chain
    *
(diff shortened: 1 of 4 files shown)


kdubb:
It was my understanding (per my memory of the 4.1 spec and the pgjdbc mailing list) that finalizers were required.  I am in the midst of replacing them with PhanomReference(s); which is a bit of a pain.  Is that not necessary? If the spec allows us to require them to properly close the object, I'll stop my changes and be happy just to merge this.

jesperpedersen:
I would say it is a 2 step solution:

1) Remove finializers
2) Add PhanomReference for allowing the auto-closing feature again

I think it is better to remove the finalizers now, since it messes up the VM. Correct JDBC behavior in all cases is a goal for the 1.0 release of course.

kdubb:
I committed the replacement for finalizers in 9db32b1beb17dc9dd3a786b8a86ecbf3308a7827. I now need to clean up the unit tests as the original authors of those didn't seem to care about leaking at all.

PR #24, opened 2013-11-01
Simplify null housekeeper contract

jesperpedersen:


kdubb:
This arguably makes the contract worse... since we always have to check for null. Is there any reason for this except for your dislike of its style?

jesperpedersen:
The "if (housekeeper != null)" check is a lot faster than invoking a noop method. I have made the variables 'final' too, such that the VM can optimize the call away completely.

Rebased against 'develop'

brettwooldridge:
Just thought I'd add a comment about JVM optimization.  If the JVM believes that there is only ever one implementation of the housekeeper, the NullHousekeeper may well be as fast as the null check.  The JVM will perform "optimistic inlining", and subsequently will "optimize away" an empty method body completely.

However, if the JVM _ever_ detects that a different implementation other than `NullHousekeeper` has been loaded, it will dynamically "de-optimize" the call sites, and the resulting `invokeinterface` will be slower than a null check.

Best demonstrated by the excellent [Mechanical Sympathy](http://mechanical-sympathy.blogspot.jp/2012/04/invoke-interface-optimisations.html) blog here.

Either will perform well, as null checks are extremely cheap, and can themselves be optimized away if the JVM hotspot compiler realizes that it has _never_ seen _housekeeper_ being null.  See slides 30-32 [here](http://www.slideshare.net/jbaruch/jvm-magic).

So, in the end they may both perform equivalently.  However, if there can ever be more than one implementation loaded at runtime, the null check will indeed be faster.

jesperpedersen:
An idea for down the road is be able to switch the housekeeper implementation at run-time through the public API of the driver

PR #27, opened 2013-11-05
Initial public API for PGConnection

jesperpedersen:
- Add .api.jdbc.PGConnection for initial public API
- Change existing classes to use internal API
- Change test cases to use either public API or internal API

A candidate for the public API could be "boolean isMinimumServerVersion(int, int)".

The DataSource related classes will provide alternative implementations of the public API.


jesperpedersen:
Rebased

jesperpedersen:
Rebased, and fix some test case leaks + added additional asserts

kdubb:
Apologies for committing with leaking tests

PR #36, opened 2013-11-14
MacAddrs and HStores codec added.

croudet:


kdubb:
Thanks for the help!

FYI, I added your new codecs to CodecTest.  In future pull requests of codecs please add an appropriate entry to this parameterized unit test.  Consequently by doing this, I found a bug in your text encoder (hex chars had 2 '6''s instead of a '6 & '7') ;)

PR #40, opened 2013-11-18
Reworked network type. Cidr added.

croudet:


kdubb:
Is there a compelling reason not to use Inet4Address & Inet6Address for the network codecs? It seems not only are these more useful java types but the parsing is already done for you as well, so less to maintain.

croudet:
Cidr accept 'partial' ipv4, i.e 10, 10.1, 10.1.2 and
they are not parsed correctly by InetAddress.getByName(inet):

10.1.2 --> /10.1.0.2 - Inet4Address
10.1 --> /10.0.0.1 - Inet4Address
10 ---> /0.0.0.10 - Inet4Address

kdubb:
Ok.  Forget parsing for the moment, is there anything precluding the codecs from returning Inet4Address & Inet6Address objects? As I said these are more useful in the Java world than our own made up types and I think we should be using them unless we absolutely cannot.

Also, with regard to parsing, it looks to me that the InetAddress classes could still be used to parse the types other than CIDR, no?

croudet:
There is not netmask field in InetAddress.
And yes InetAddress could be used to parse ip.

kdubb:
Ok. I see now.  I hadn't realized that we needed to specifically carry the net mask around as well.  These seem fine with me then.

kdubb:
I merged this in but with numerous simplifications on the parsing for Inet and Cidr.  I also changed names to be InetAddr and CidrAddr because Inet is way to general in my opinion (and changed Cidr to match).

Thx for the work!

PR #41, opened 2013-11-20
Add OSGi headers to MANIFEST.MF

brettwooldridge:
Add OSGi headers to MANIFEST.MF so that the JAR can be used as an OSGi bundle in containers like Apache Felix and Eclipse Equinox.


kdubb:
Is it possible to export only a single class in OSGi? I'm thinking it would be better to export PGDriver only.

Either way, the _com.impossibl.postgres.api.jdbc_ package needs to be exported as well.

brettwooldridge:
In OSGi It is not possible to export a single class.  Basically, any package with classes that the user can directly reference (by class name) needs to be exported.  Will a user directly reference PGConnection in their code?  If so, I'll add _com.impossibl.postgres.api.jdbc_ package to the exports.

kdubb:
Yes.  The c.i.p.jdbc.api package is for the public API of the driver specific extensions to the JDBC API.

brettwooldridge:
Added.

PR #48, opened 2013-12-02
use Deque instead of Stack in SQLText (avoid synchronized methods.) …  Remove unnecessary toString in Timestamps.

croudet:


kdubb:
The SQL text changes actually break all the unit tests.  I thought I had tested the PR before merging but was testing on another branch.

Are you running all tests before making a PR?

croudet:
Sorry about that forgot to push the fix:
in public static MultiStatementNode parse(String sql)
it should be
return (MultiStatementNode)parents.getLast();
and not getFirst()

PR #50, opened 2013-12-04
Change jdbc:postgresql to jdbc:pgsql

jesperpedersen:
The original driver uses jdbc:postgresql and DriverManager in order to create connections. This can lead to problems if both the original driver and pgjdbc-ng are deployed.

Hence change the URL to jdbc:pgsql in order to not cause collisions.

DriverManager is really a broken concept :/

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,11 +8,11 @@ A new JDBC driver for PostgreSQL aimed at supporting the advanced features of JD
 
 The connection format for the pgjdbc-ng driver is
 
-	jdbc:postgresql://<server>[:<port>]/<database>
+	jdbc:pgsql://<server>[:<port>]/<database>
 
 An example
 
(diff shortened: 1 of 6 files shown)


jesperpedersen:
Rebased

kdubb:
Can we definitively say that allowing the use of both drivers in the same JVM is better than providing a drop in replacement?

jesperpedersen:
Well, if it is drop-in then we need to support all the connection parameters, use the same class names for DataSource, XADataSource, and the same public API.

IMHO, it isn't a drop-in - it is a new implementation with its own parameters and API.

Of course it would be different if this implementation completely replaced the official driver on jdbc.postgresql.org. So atm, I think it is safest to have its own identifier

kdubb:
You convinced me ;)

PR #53, opened 2013-12-06
OSGi manifest header changes

brettwooldridge:
Add XA package to OSGi manifest headers.  Also generate OSGI headers for the default JAR, we were only generating OSGi headers for the complete JAR.


brettwooldridge:
When the port to Netty 4 is finished, there will be a minor change needed to the default (non-complete) jar to add Import-Package headers...

PR #54, opened 2013-12-06
Geomtric types added

croudet:
Geometric types added:
     - Points
     - Line Segments
     - Boxes
     - Paths
     - Polygons
     - Circles


croudet:
All test passed!

Results :

Tests run: 754, Failures: 0, Errors: 0, Skipped: 6

src/main/java/com/impossibl/postgres/protocol/v30/ProtocolShared.java
```diff
@@ -112,7 +112,7 @@ private void init() {
     Executor bossExecutorService = Executors.newCachedThreadPool(new NamedThreadFactory("PG-JDBC Boss"));
     Executor workerExecutorService = Executors.newCachedThreadPool(new NamedThreadFactory("PG-JDBC Worker"));
 
-    int workerCount = getRuntime().availableProcessors();
+    int workerCount = Math.max(10, getRuntime().availableProcessors());
```
kdubb: Please remove this change.  If anything this max should be a connection setting.

src/test/java/com/impossibl/postgres/jdbc/ParameterMetaDataTest.java
```diff
@@ -74,7 +74,7 @@ public void testParameterMD() throws SQLException {
     assertEquals(Types.VARCHAR, pmd.getParameterType(2));
     assertEquals("text", pmd.getParameterTypeName(2));
     assertEquals("java.lang.String", pmd.getParameterClassName(2));
-    assertEquals(Types.OTHER, pmd.getParameterType(3));
+    assertEquals(Types.BINARY, pmd.getParameterType(3));
```
kdubb: What is the reason for this change?

jesperpedersen:
Shouldn't we consider some of this for .api. ?

jesperpedersen:
There is some code that commented out, remove that

jesperpedersen:
Using "git rebase -i upstream/develop" to pull in upstream changes makes it easier to merge

kdubb:
@jesperpedersen, 

@croudet is just following my lead.  I had originally put all the system data type classes (e.g. ACLItem) in .data. (for some unknown reason lol).  All of the classes in .data. should be moved to .api. at some point... I can do that once this is merged

jesperpedersen:
@kdubb sounds good :)

src/test/java/com/impossibl/postgres/jdbc/ParameterMetaDataTest.java
croudet: I am not sure what type to put for getInputPrimitiveType() and getOutputPrimitiveType(). Currently it is Binary, should it be Array?

kdubb: You can try to use Unknown.

Since these are technically "system" types, unrecognized by JDBC, you could go the long route and add an enum entry for each of them to PrimitiveType. Although, that change cascades into changes in SQLTypeMetadata, SQLTypeUtils, etc.

kdubb:
@croudet  please do rebase as @jesperpedersen has shown when updating your PR branches.  It makes the commit history much nicer.

jesperpedersen:
@kdubb and you shouldn't do merge commits ;) lol

kdubb:
@jesperpedersen If I am being honest... I am no git assassin lol   I just noticed your PR's seem to come in as just the commits themselves.  I need to do some learning I know ;)

jesperpedersen:
@kdubb see the manual instructions on the PR for starters, you can always rebase in the master repo -- the key is "git merge --ff-only"

src/test/java/com/impossibl/postgres/jdbc/ParameterMetaDataTest.java
croudet: When I use Unknown I got "type has no supported parameter format" error (type.getParameterFormat().)

croudet:
I have added the PrimitiveTypes for the geometric types.

kdubb:
Ok so this looks pretty good.  One question I now have is,  why no types for Point, Line, Circle, etc?

Good work though!

croudet:
I try to use types as 'raw' as possible so that the user can map them to what ever he wants.
For Path, you need to store the open/close flag so I created a class.
If you think it more convenient to have classes, I can create classes.

kdubb:
I'm fine with it the way it is... in the future I think adding specific classes for these types would be better.

kdubb:
Merged (by rebasing)

PR #55, opened 2013-12-09
Close any previous generated keys ResultSet before assigning a new value

jesperpedersen:


src/main/java/com/impossibl/postgres/jdbc/PGStatement.java
```diff
@@ -227,6 +228,11 @@ static void closeResultSets(List<WeakReference<PGResultSet>> resultSets) {
    */
   void closeResultSets() throws SQLException {
     closeResultSets(activeResultSets);
+
```
kdubb, line 231: This is unnecessary as generatedKeysResultSet is added to activeResultSets when created

src/main/java/com/impossibl/postgres/jdbc/PGPreparedStatement.java
```diff
@@ -208,6 +208,10 @@ public boolean execute() throws SQLException {
     }
 
     if (wantsGeneratedKeys) {
+      if (generatedKeysResultSet != null && !generatedKeysResultSet.isClosed()) {
+        generatedKeysResultSet.internalClose();
+      }
+
```
kdubb: This needs to remove generatedKeysResultSet from the activeResultSets list as well.

src/main/java/com/impossibl/postgres/jdbc/PGPreparedStatement.java
```diff
@@ -318,6 +322,10 @@ public void clearBatch() throws SQLException {
 
       }
 
+      if (generatedKeysResultSet != null && !generatedKeysResultSet.isClosed()) {
+        generatedKeysResultSet.internalClose();
+      }
+
```
kdubb: This code should be removed.  executeBatch is missing a call to closeResultSets, which would close all activeResultSets, at the outset of the method; which all executeXXX methods are required to do.  generatedKeysResultSet is in the activeResultSets list.

jesperpedersen:
Fixed

PR #60, opened 2014-01-20
Much more efficient statement name generation.

brettwooldridge:
This is a trivial change, but in a test in which I was creating 100,000 prepared statements, the `getNextStatementName` went from consuming 8.8% of the measured time to not even being registered by the profiler.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,6 +64,11 @@
 			<artifactId>netty</artifactId>
 			<version>3.8.0.Final</version>
 		</dependency>
+        <dependency>
+        	<groupId>com.googlecode.concurrentlinkedhashmap</groupId>
+        	<artifactId>concurrentlinkedhashmap-lru</artifactId>
+        	<version>1.4</version>
+        </dependency>
 	</dependencies>
 
 	<build>
--- a/src/main/java/com/impossibl/postgres/jdbc/PGConnectionImpl.java
+++ b/src/main/java/com/impossibl/postgres/jdbc/PGConnectionImpl.java
@@ -270,9 +270,7 @@ public class PGConnectionImpl extends BasicContext implements PGConnection {
    * @return New unique statement name
    */
   String getNextStatementName() {
-    StringBuilder sb = new StringBuilder("0000000000000000");
-    sb.append(Long.toHexString(++statementId));
-    return sb.substring(sb.length() - 16);
+    return Long.toHexString(++statementId);
   }
 
   /**
@@ -281,7 +279,7 @@ public class PGConnectionImpl extends BasicContext implements PGConnection {
    * @return New unique portal name
    */
   String getNextPortalName() {
-    return String.format("%016X", ++portalId);
+    return Long.toHexString(++portalId);
   }
 
   /**
--- a/src/main/java/com/impossibl/postgres/jdbc/PGConnectionImpl.java
+++ b/src/main/java/com/impossibl/postgres/jdbc/PGConnectionImpl.java
@@ -95,6 +95,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.Executor;
 
 import static java.lang.Boolean.parseBoolean;
@@ -105,6 +106,8 @@ import static java.sql.Statement.RETURN_GENERATED_KEYS;
 import static java.util.Arrays.asList;
 import static java.util.Collections.unmodifiableMap;
 import static java.util.concurrent.TimeUnit.SECONDS;
+
+import com.googlecode.concurrentlinkedhashmap.ConcurrentLinkedHashMap;
 
 import org.jboss.netty.handler.queue.BlockingReadTimeoutException;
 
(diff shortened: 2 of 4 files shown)


jesperpedersen:
Could you change getNextPortalName in the same way so they "match" ?

kdubb:
Also, as I have no idea why I wanted it padding to 16 zeros; it an astronomical max for this field and completely useless padding.  Just go all the way and forgo the padding altogether and return Long.toHexString result.

brettwooldridge:
Will do.  On the way into the office now (8am here in Tokyo).

brettwooldridge:
Wah, what did github do?  I submitted a second, unrelated pull request, but github apparently merges them together.

The "second" pull request (d75f05b) is more for code review than integration.  The change adds a parsed SQL cache to `PGConnectionImpl`.  In my testing, with 10,000 prepare statement calls, it reduces the overall time from _2388ms_ to _768ms_ (> 3x).  It also has the effect of reducing GC pressure by over 90%.  The statement that was tested was fairly complex to rigorously test the change:

``` sql
SELECT ZDeviceLite.device_id,
       ZDeviceLite.ip_address,
       ZDeviceLite.ip_high,
       ZDeviceLite.ip_low,
       ZDeviceLite.hostname,
       ZDeviceLite.network,
       ZDeviceLite.adapter_id,
       ZDeviceLite.device_type,
       ZDeviceLite.vendor_hw,
       ZDeviceLite.model,
       ZDeviceLite.vendor_sw,
       ZDeviceLite.canonical_os_version,
       ZDeviceLite.os_version,
       ZDeviceLite.asset_identity,
       ZDeviceLite.backup_status,
       ZDeviceLite.compliance_state,
       ZDeviceLite.last_backup,
       ZDeviceLite.last_telemetry,
       ZDeviceLite.memo_summary,
       ZDeviceLite.custom1,
       ZDeviceLite.custom2,
       ZDeviceLite.custom3,
       ZDeviceLite.custom4,
       ZDeviceLite.custom5
FROM device ZDeviceLite
JOIN
  (SELECT t0.device_id
   FROM device t0
   JOIN
     (SELECT t1.device_id
      FROM device t1 JOIN
        (SELECT device_id
         FROM device d
         WHERE d.ip_low BETWEEN 167772672 AND 167772927
           AND network='Default'
         UNION SELECT d2.device_id
         FROM device d2,
           (SELECT i.device_id
            FROM device_interface_ips i
            WHERE i.ip_low BETWEEN 167772672 AND 167772927) baz
         WHERE d2.device_id=baz.device_id
           AND d2.network='Default') AS gfd ON t1.device_id=gfd.device_id) AS j1 ON t0.device_id=j1.device_id
   AND t0.vendor_hw=?) AS j0 ON ZDeviceLite.device_id=j0.device_id
AND ZDeviceLite.network=?
ORDER BY ZDeviceLite.ip_high ASC,
         ZDeviceLite.ip_low ASC
OFFSET 0 ROWS FETCH NEXT 254 ROWS ONLY
```

Items of note:
- Currently the cache is fixed size (500).  This should probably be configured through a property (eg. maxCachedParsedSQL).
- Currently it adds an additional maven dependency for Google's ConcurrentLinkedHashMap.  The ConcurrentLinkedHashMap has been integrated into Guava, and I noticed we have a Guava sub-package where I assume classes from Guava have been dropped.  If we want to eliminate the maven dependency, I can place the code into the guava sub-package (it has dependencies that reach 10 classes or so).

kdubb:
I believe you'll need to add the parameter types to the statement cache key. Postgres take the types into account during the bind (aka prepare).

To test this run a query such as

``` sql
SELECT * FROM test WHERE value = ?
```

using a string parameter then using an integer parameter.  I am imagining in the current incarnation this would fail.

brettwooldridge:
This is not the (hopefully forthcoming) change to add a prepared statement cache.  That will indeed require a more complex cache key.

This change works by caching the _initial_ parse (`SQLText`), and then handing out _deep-copies_ of the `SQLText` instance for each call to `PGConnectionImpl.parseSQL(String)`.  Because the initial parse is parameter type agnostic, there is no issue.  Each `PGPreparedStatement` (or `PGCallableStatement`) gets its own _unique_ copy of the `SQLText` to modify as they see fit.  From the perspective of everything else in the code (that consumes `parseSQL(String)`) it behaves as if a fresh parse was performed.

On an unrelated note, the more I think about it the more I think that the `ConcurrentLinkedHashMap` dependency can be replaced with a simple `Collections.synchronizedMap(new HashMap<>)` without any measurable impact on performance.

kdubb:
My apologies I guess I wasn't paying attention; even know I had read through everything.  I'm now astonished that you have brought it's performance so close to that of the classic driver with such a fairly simple caching scheme.  Good work.

With regards to the use of guava classes... Yes, drop then into our own package.  The shade plugin was including way too many extraneous classes and leaving the jar too bloated.

brettwooldridge:
I went ahead and replaced the `ConcurrentLinkedHashMap` with a `Collections.synchronizedMap(new LinkedHashMap<>())`.  The lock only exists for the duration of a `map.get()` or `map.put()` call which shouldn't cause any real blocking unless hundreds of threads are preparing statements simultaneously, which seems unlikely.

Just a note on the pgjdbc-ng vs. the classic driver, the "scale" of this test was different -- 10,000 iterations, not 100,000.  I haven't measured the classic vs. pgjdbc-ng after this addition yet.  But as you noted, the classic driver isn't really taking a roundtrip to the server in a simple prepare case.

jesperpedersen:
@brettwooldridge Any chance that you can split this PR up into 2 separate ones making it easier to look at your changes ? I wrote a wiki page here: https://github.com/impossibl/pgjdbc-ng/wiki/pgjdbc-ng-development

brettwooldridge:
I'll delete this pull request (if I can) and resubmit.

brettwooldridge:
Closing to resubmit a cleaner pull.

PR #61, opened 2014-01-21
Introduce a parsed SQL cache for SQLText hierarchies.

brettwooldridge:
Ok this is a freshened-up pull request.  I added the ability to configure the cache size and removed the dependency on Google ConcurrentLinkedHashMap.  Note also that this does not use Java serialization, it uses custom code to perform the SQLText deep-copy which performs almost an order of magnitude faster than standard (notoriously slow) Java serialization.

Unlike the (forthcoming) prepared statement cache which will default to 'off', the parsed SQL cache defaults to 'on' because it not only improves performance, with no side-effects, but also reduces GC pressure which we should view as a good thing.  One of the primary goals of Netty4 was reducing GC pressure, because in a long-running high-performance system it starts to become an issue.

### DIFF
--- a/src/main/java/com/impossibl/postgres/jdbc/AbstractDataSource.java
+++ b/src/main/java/com/impossibl/postgres/jdbc/AbstractDataSource.java
@@ -29,6 +29,7 @@
 package com.impossibl.postgres.jdbc;
 
 import com.impossibl.postgres.system.Context;
+import com.impossibl.postgres.system.Settings;
 
 import java.io.PrintWriter;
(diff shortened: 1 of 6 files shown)


jesperpedersen:
You can look into if you want to expose it as a parameter for the DataSource implementations too

brettwooldridge:
Ok, I have made the suggested changes:
- parsedSqlCacheSize <= 0 disables the cache
- cache size is configurable on the DataSource

I have squashed those commits, so I am going to close this pull request and resubmit the final.

Issue #58, opened 2014-01-20
Implement driver-side prepared statement cache


brettwooldridge:
Every major database vendor JDBC driver provides a prepared statement cache...except the classic PostgreSQL driver.  For example:

[MySQL](http://dev.mysql.com/doc/connector-j/en/connector-j-reference-configuration-properties.html)'s DataSource.  **Defaults to caching 25 statements.**

```
MysqlDataSource.setPrepStmtCacheSize()
```

[Oracle](http://docs.oracle.com/cd/B10500_01/java.920/a96654/stmtcach.htm) has both _implicit_ and _explicit_ caching (see link).

```
OracleDatasource.setMaxStatements();
```

[DB2](http://publib.boulder.ibm.com/infocenter/db2luw/v9r5/index.jsp?topic=%2Fcom.ibm.db2.luw.wn.doc%2Fdoc%2Fc0052011.html).

```
"maxStatements specifies the maximum number of statements that
 can be kept open by the driver in an internal statement cache"
```

[SQL Server](http://jtds.sourceforge.net/faq.html) via jTDS:

```
cacheMetaData (default - false)
```

As one of the authors of the relatively new [HikariCP](https://github.com/brettwooldridge/HikariCP) connection pool, we took the decision _not_ to implement statement caching in the pool because it is so widely implemented by JDBC drivers.  Not only does it add complexity to the pool, but JDBC drivers are actually in a better position to judge for a given database, given configuration, and given statement whether to cache or not cache.

In light of this, I think it would be great of pgjdbc-ng made up for the failings of "classic" PostgreSQL driver by offering prepared statement caching.

I am willing to implement, if pgjdbc-ng agrees in principal that it is useful.

kdubb:
So that I am clear on understanding, please clarify how the statements would be cached.  I imagine you are talking about caching statements based on the SQL text itself?

brettwooldridge:
Yes, I would envision caching the PGPreparedStatement instances based on the raw SQL text.  A "close" in this case would return the statement to the cache (an LRU cache).  If the connection is closed all statements in the LRU cache would be closed.  If adding a statement to the LRU pushes a statement out of the cache (capacity exceeded), the evicted statement would be closed.

[...]

PR #62, opened 2014-01-23
Introduce a parsed SQL cache for SQLText hierarchies.

brettwooldridge:
Description copied from original request:

```
The change adds a parsed SQL cache to PGConnectionImpl. In my testing, with 10,000
prepare statement calls, it reduces the overall time from 2388ms to 768ms (> 3x). It also
has the effect of reducing GC pressure by over 90%.
```

 Contains recommended changes.

### DIFF
--- a/src/main/java/com/impossibl/postgres/jdbc/AbstractDataSource.java
+++ b/src/main/java/com/impossibl/postgres/jdbc/AbstractDataSource.java
@@ -29,6 +29,7 @@
 package com.impossibl.postgres.jdbc;
 
 import com.impossibl.postgres.system.Context;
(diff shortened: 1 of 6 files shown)


brettwooldridge:
I have fixed work on #58 (Implement driver-side prepared statement cache), but it is pending on this pull as it touches many of the same files.  I can submit that pull for review as soon as this pull request is resolved.

brettwooldridge:
Do you need me to rebase this on the Netty4 commit, or will you handle it on your side?

kdubb:
That would be helpful

brettwooldridge:
Rebased.

kdubb:
Merged... Thx for the work!

PR #64, opened 2014-01-24
Implement prepared statement caching.

brettwooldridge:
This is a pull request that implements #58 (Implement driver-side prepared statement cache).

The largest block of code below is the class `LruStatementCache`.  This class is a _modified_ version of a prepared statement cache that I wrote for the [Bitronix JTA Transaction Manager](https://github.com/bitronix/btm).  The BTM original class is [here](https://github.com/bitronix/btm/blob/master/btm/src/main/java/bitronix/tm/resource/jdbc/LruStatementCache.java) and has been in use for several years, so it's pretty solid (no reported bugs).  It is Apache Licensed, and I wrote it, so there should be no issue using it.

The performance numbers for this change are as follows for 100,000 prepare statements:

```
Original                         : 27945ms
Original+parsed SQL cache (PSC)  : 11721ms
Original+PSC+statement cache     : 56ms
```

<sup>1</sup> The [parsed SQL cache](https://github.com/impossibl/pgjdbc-ng/pull/62) was recently pulled.

The prepared statement cache is per-connection.  PGPreparedStatements are "single-use", by which I mean ... when a statement is prepared, if it exists in the cache it is removed from the cache and re-used, during which time if the client re-prepares the same statement it is _not_ found in the cache and a new PGPreparedStatement will be created.  When a PGPreparedStatement is closed, it is returned to the cache.  The cache is this case is acting more like a "pool" (a "pool" is technically still a "cache").

The next biggest change after the inclusion of `LruStatementCache` are the changes to `PGConnectionImpl`. The `PGConnectionImpl.java` diffs below are a little difficult to read as rendered, but in fact are rather simple in nature.  I suggest opening `PGConnectionImpl.java` in a separate browser window and following the diffs along in the full source for readability.

All tests are passing with both the cache turned on and with the cache turned off.  There are two changes to unit tests included here which were leaking statements and causing noise in the test output, so I added close() calls to those tests' statements.

### DIFF
--- a/src/main/java/com/impossibl/postgres/jdbc/AbstractDataSource.java
+++ b/src/main/java/com/impossibl/postgres/jdbc/AbstractDataSource.java
@@ -53,6 +53,7 @@ public abstract class AbstractDataSource implements CommonDataSource {
   private String password;
   private boolean housekeeper;
   private int parsedSqlCache;
+  private int prepStmtCacheSize;
 
   /**
    * Constructor
@@ -66,6 +67,7 @@ public abstract class AbstractDataSource implements CommonDataSource {
     this.password = null;
     this.housekeeper = true;
     this.parsedSqlCache = 250;
+    this.prepStmtCacheSize = 0;
   }
 
   /**
@@ -218,6 +220,23 @@ public abstract class AbstractDataSource implements CommonDataSource {
     parsedSqlCache = cacheSize;
   }
 
+  /*
+   * Get the size of the PreparedStatment cache.
+   * @return the maximum number of PreparedStatements cached per connection
+   */
+  public int getPreparedStatementCacheSize() {
+    return prepStmtCacheSize;
+  }
+
+  /**
+   * Set the size of the PreparedStatment cache. A value of 0 will disable
+   * the cache.
+   * @return the maximum number of PreparedStatements cached per connection
+   */
+  public void setPreparedStatementCacheSize(int cacheSize) {
+    this.prepStmtCacheSize = cacheSize;
+  }
+
   /**
    * Create a connection
    * @param u The user name
@@ -254,6 +273,7 @@ public abstract class AbstractDataSource implements CommonDataSource {
       hk = new ThreadedHousekeeper();
 
     props.put(Settings.PARSED_SQL_CACHE, parsedSqlCache);
+    props.put(Settings.PREPARED_STATEMENT_CACHE, prepStmtCacheSize);
 
     return ConnectionUtil.createConnection(url, props, hk);
   }
--- a/src/main/java/com/impossibl/postgres/jdbc/LruStatementCache.java
+++ b/src/main/java/com/impossibl/postgres/jdbc/LruStatementCache.java
@@ -1,0 +1,232 @@
+/**
+ * Copyright (c) 2013, impossibl.com
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *  * Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *  * Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *  * Neither the name of impossibl.com nor the names of its contributors may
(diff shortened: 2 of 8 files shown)


kdubb:
Holy cow.... cannot argue with those #'s.  Give me some time to digest this.

brettwooldridge:
Feel free to annotate with questions.

brettwooldridge:
Changes made and squashed.

jesperpedersen:
Looks good

kdubb:
After looking at this in depth.  I am in belief that this could be done with less structural changes by doing it at a bit lower level.  For example, in PGPreparedStatement.parseIfNeeded simply hashing the SQLText and parameter types, using that hash as the actual server statement name and then chucking those names into a LRU cache seems to do the trick.  Eviction, and cleanup, could be elegantly handled through that cache which means it touches no other code.

I hate to argue with success but there does seem to be a simpler way to achieve this.  Am I missing something?

brettwooldridge:
What about things like cursor holdability or concurrency?  Just because two SQL statements are prepared with the same SQL text and parameter types does not make the statements equivalent, because one can be prepared with held cursors, or repeatable_read, for example, and the other not.

kdubb:
Actually, in terms of Postgres's FE/BE protocol, all prepared statements are equal.   All the other stuff you reference does not affect the prepared statement.  You can specifically prepare a statement and execute it, change the isolation level, and execute it again with the new isolation level.

If I am wrong on this... anything to the contrary is certainly does not seem to be documented anywhere in Postgres.

brettwooldridge:
The cache at the higher-level does avoid manipulation of the SQLText.  For example, when you prepare a statement with auto-generated keys, additional nodes are added to the SQLText tree (RETURNING...) ... and that along with the PGPreparedStatement are cached.  If we cache at the lower level, we still have to go through the SQLText manipulation, and after that determine equivalency of hashes.  I could be wrong, but I would expect that to fall closer to the 11000ms benchmark than the higher level cache of 56ms.  But only doing it and benchmarking it would tell.

brettwooldridge:
After thinking about it, you might be right.  While I think the performance will be lower than caching at a higher-level, the trade-off for simplicity might be worth it.  Have you be able to test the idea?  If so, can I pull it from somewhere and try it out?

kdubb:
I am working on it as we speak... I'll let you know my results soon

kdubb:
Can you attach the test you used to generate those performance #'s? I want to compare apples to apples.

brettwooldridge:
I'm on the road in Europe so I don't have access to my system back in Tokyo, but the basics of it was just:

```
long start = System.nanoTime();
for (int i = 0; i < 10000; i++) {
    PreparedStatement prepareStatement = connection.prepareStatement("");
    prepareStatement.close();
}
System.out.println("Prepare: " + TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start));
```

Where the prepared statement text was this: [https://gist.github.com/brettwooldridge/8668029](https://gist.github.com/brettwooldridge/8668029)

That is a query against our in-house application, so I don't have a simple test case I can bottle up for you.  But any sufficiently complex query with a couple of replacement parameters would serve well as a test, I think.

brettwooldridge:
Any progress on your idea for the low-level caching strategy vs. the high-level one proposed?

kdubb:
I made some progress and it looks promising... unfortunately I have been buried with a new project for a while.

kdubb:
@brettwooldridge  I have working branch that's passing all tests and with your given 10000 iteration test the run takes ~1 second.  It's not 56ms but my implementation changes basically 2 files and isolated to 3 methods. Also, the test you provided is a bit contrived since the statement is never being executed.  Considering that I think I prefer the simpler implementation but I'd like to know your thoughts.

My implementation is available in the "cached-statements" branch at https://github.com/impossibl/pgjdbc-ng/tree/cached-statements

This branch also contains your test in the PerformanceTests unit test.

P.S. Currently my implementation is using a bunch of the caching classes from Guava and thus includes the full guava in the pom.  I have to replace or transplant these classes before it can be merged mainline.

kdubb:
Somehow I managed to miss the part where you said to insert a complex query with replacement parameters into your tests.

After doing this the it raised my runtime to ~3 seconds.  Unfortunately I cannot compare against your #'s since the query itself comes into play.

Please check this out with your query and update me.

kdubb:
FYI... I readied the "cached-statements" branch for merging with mainline.

PR #69, opened 2014-02-22
Implement prepared statement caching (alternative approach)

kdubb:
This pull request implements #58 in a simpler fashion than #64

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,6 +64,11 @@
 			<artifactId>netty-all</artifactId>
 			<version>4.0.15.Final</version>
 		</dependency>
+		<dependency>
+			<groupId>com.google.guava</groupId>
+			<artifactId>guava</artifactId>
+			<version>16.0</version>
+		</dependency>
 	</dependencies>
 
 	<build>
--- a/pom.xml
+++ b/pom.xml
@@ -64,11 +64,6 @@
 			<artifactId>netty-all</artifactId>
 			<version>4.0.15.Final</version>
 		</dependency>
-		<dependency>
-			<groupId>com.google.guava</groupId>
-			<artifactId>guava</artifactId>
-			<version>16.0</version>
-		</dependency>
 	</dependencies>
 
 	<build>
--- a/src/main/java/com/impossibl/postgres/jdbc/AbstractDataSource.java
+++ b/src/main/java/com/impossibl/postgres/jdbc/AbstractDataSource.java
@@ -53,6 +53,7 @@ public abstract class AbstractDataSource implements CommonDataSource {
   private String password;
   private boolean housekeeper;
   private int parsedSqlCache;
+  private int preparedStatementCacheSize;
 
   /**
    * Constructor
@@ -66,11 +67,13 @@ public abstract class AbstractDataSource implements CommonDataSource {
     this.password = null;
     this.housekeeper = true;
     this.parsedSqlCache = 250;
+    this.preparedStatementCacheSize = 0;
   }
 
   /**
    * {@inheritDoc}
    */
+  @Override
   public int getLoginTimeout() throws SQLException {
     return loginTimeout;
   }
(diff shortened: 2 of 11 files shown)


brettwooldridge:
I have posted a self-contained gist that contains the query that I was using to test and creates the necessary tables to execute the query.  The gist is ready to run.

https://gist.github.com/brettwooldridge/9180671

Using this query against my local PostgreSQL server, the performance of the two approaches comes in at:

```
Original..... 75ms
Alternative.. 748ms
No-cache..... ~2200ms
```

kdubb, I appreciate the effort to propose an alternative, and not to overly defend my original implementation, but I question whether the alternative patch can be classified as "simpler".  The alternative touches the same core set of classes, contains +163 changed lines vs. +93 in the original, and introduces +1172 lines of new dependencies vs. +232 in the original.  The total patch size is +325 lines for the original vs. +1335 lines in the alternative.  The alternative is also almost exactly 1 order of magnitude slower.

Original:

```
AbstractDataSource.java    +20
PGConnectionImpl.java      +65/-20
PGPreparedStatement.java   +11/-2
PGStatement.java           +10/-2
Settings.java              +1
LruStatementCache.java     +232    (new)

Total: +325 lines
Excluding new files: +93 lines
```

Alternative:

```
AbstractDataSource.java    +34/-3
PGConnectionImpl.java      +84
PGPreparedStatement        +45/-5
PGStatement                +4
Settings.java              +4
Cache                      +152    (new)
CacheBuilder               +870    (new)
RemovalListener            +48     (new)
RemovalNotification        +102    (new)

Total: +1335 lines
Excluding new files: +163
```

Even excluding the 1172 lines of dependencies from guava, the total changes lines are greater in the proposed alternative.  Note the "housekeeper" changes were excluded from the above counts.

brettwooldridge:
I'm also not sure what is up with the alternative.  In my test, it doesn't matter whether `setPreparedStatementCacheSize()` is `0` or `50`, the benchmark results are the same.  Is the cache being used irrespective of the setting?

kdubb:
@brettwooldridge You must be comparing this with the wrong commit.  I removed all the google guava usage from the code.   I have made a few changes in here besides those directly affecting statement caching and made it a little confusing.  Removing all these extraneous changes and I believe my changes will be less code but I'd have to count it.

I have two points to make about "simpler"... 

1) regardless of the LOC between the two only 1 method was changed in PGStatement and 1 method in PGConnectionImpl. Although, some utility functions were added in both these don't change the way the current code works.

2) In your implementation you have taken a fundamentally unshared object (PGPreparedStatement) and started sharing it.  Knowing the complexity of these classes (well the JDBC spec really) this change scares me simply because I didn't write the code this way and don't really want to have to worry about it in the future. This is the crux of my "simpler" argument.  In my implementation everything works exactly as it used.

WRT setPreparedStatementCacheSize not working... I have never tested it using it from a Datasource. I'll do that now.

kdubb:
I see the difference... you are just creating the prepared statement and closing it, never executing it.  

My implementation will see no difference unless the statement is actually executed.  The difference you are seeing here is because in your implementation you are skipping the SQL processing and I am not.

brettwooldridge:
Understand re: simplicity.  Including execution of the statement, the delta between the two proposals remains roughly the same, but essentially becomes "noise" in the measurement (15057ms vs 15690ms).  Both are ~8000ms faster than no-cache.  I rescind any concerns over the alternative (save for the aforementioned 0 cache size issue).

kdubb:
I have confirmed the setPreparedStatementCache(0) does disable the cache correctly.  Again, as previously stated, in your original gist you never called execute on the PreparedStatement.  My code will never kick-in unless you call execute.

With that confirmed, it sounds like we are in agreement to merge this mainline?

brettwooldridge:
Yes, go ahead and merge.  Thanks for the work.

kdubb:
@brettwooldridge Thank you for the work you did.  It probably wouldn't have gotten completed without you starting it off.

PR #73, opened 2014-03-29
Fix connecting to devel version of postgres

gmokki:
The current code throws NumberFormatException when trying to parse git checkout postgres server version '9.4devel'.
While at it fix the code to not eat the above mentioned exception when connecting.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,7 +71,7 @@
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-compiler-plugin</artifactId>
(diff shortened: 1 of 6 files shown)


gsmet:
+1

Note that it's going to conflict with https://github.com/impossibl/pgjdbc-ng/pull/74 but it's an easy fix.

brettwooldridge:
@gmokki pull request #74 has been merged, which creates a compiled regular expression constant for version splitting.  Can you resubmit this pull request with your changes integrated?

kdubb:
This was combined into pr #91

PR #74, opened 2014-04-08
Compile the version split pattern.

gsmet:
It's usually considered a best practice even if it's not a hot spot.


brettwooldridge:
This pull request looks reasonable, provides a minor performance improvement, and makes a constant string more visible for maintainability.

PR #75, opened 2014-04-08
implements Interval coercion

gsmet:
Interval coercion is missing and it's an issue for one of our applications.

I implemented it as everything was already ready for that and added a test case.

### DIFF
--- a/src/main/java/com/impossibl/postgres/jdbc/SQLTypeUtils.java
+++ b/src/main/java/com/impossibl/postgres/jdbc/SQLTypeUtils.java
@@ -28,6 +28,7 @@
  */
 package com.impossibl.postgres.jdbc;
 
+import com.impossibl.postgres.api.data.Interval;
 import com.impossibl.postgres.api.data.Path;
 import com.impossibl.postgres.api.data.Record;
 import com.impossibl.postgres.api.data.Tid;
@@ -227,6 +228,9 @@ class SQLTypeUtils {
     }
     else if (targetType == Instant.class) {
       return coerceToInstant(val, sourceType, zone, connection);
+    }
+    else if (targetType == Interval.class) {
+      return coerceToInterval(val);
     }
     else if (targetType == URL.class) {
       return coerceToURL(val);
@@ -542,6 +546,9 @@ class SQLTypeUtils {
     else if (val instanceof Timestamp) {
       return val.toString();
     }
+    else if (val instanceof Interval) {
+      return val.toString();
+    }
     else if (val instanceof Instant) {
       return ((Instant) val).disambiguate(TimeZone.getDefault()).print(context);
     }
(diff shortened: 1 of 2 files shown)


brettwooldridge:
@gsmet This looks good, can you add a test for coercion failure?  And if possible, add a test for _interval_ to _string_ conversion?

gsmet:
Hi Brett,

Any pointers on similar tests for the other coercions? IIRC, I tried to be consistent with the tests existing for the other coercions but I might have missed something and, if so, I would prefer to take a look at what's existing to be as much as similar as possible.

Thanks.

brettwooldridge:
To be honest, I haven't looked at the other (existing) coercion tests and it is possible that they are less than thorough.  I was just basing my recommendations on looking at the changes you made as well as the test to cover them.  It seemed that only the "positive" coercion case was tested, and only going from String→Interval.  I leave it to your discretion as to how deeply to test, but it seems that adding two simple tests, one coercion failure, and one Interval→string, would be enough.

gsmet:
OK, I'll see what I can do.

I'm on the road this week so don't hold your breath but I'll take a shot at it next week so that we can get this committed soon.

gsmet:
I had some spare time so I polished the tests a bit.

I'm not sure it's a good idea to depend on the implementation details of the Interval string representation.

Thoughts?

brettwooldridge:
Is the string representation something that can change, i.e. configurable in PostgreSQL in some way?  Or is it entirely internal to the pgjdbc-ng driver?  If a change in the Interval string representation could break pgjdbc-ng's handling of the Interval, I think it is acceptable (and valid) to test based on the implementation details.

gsmet:
> Is the string representation something that can change, i.e. configurable in PostgreSQL in some way?

The interval style is configurable in PostgreSQL: there are several variants.

See:
- http://www.postgresql.org/docs/9.1/static/runtime-config-client.html#GUC-INTERVALSTYLE
- http://www.postgresql.org/docs/9.1/static/datatype-datetime.html#DATATYPE-INTERVAL-INPUT

But in our case, it should be stable from the driver point of view as it treats the interval values in a non ambiguous way.

See especially this paragraph:

> According to the SQL standard all fields of an interval value must have the same sign, so a leading negative sign applies to all fields; for example the negative sign in the interval literal '-1 2:03:04' applies to both the days and hour/minute/second parts. PostgreSQL allows the fields to have different signs, and traditionally treats each field in the textual representation as independently signed, so that the hour/minute/second part is considered positive in this example. If IntervalStyle is set to sql_standard then a leading sign is considered to apply to all fields (but only if no additional signs appear). Otherwise the traditional PostgreSQL interpretation is used. To avoid ambiguity, it's recommended to attach an explicit sign to each field if any field is negative.

As for the other point:

>  If a change in the Interval string representation could break pgjdbc-ng's handling of the Interval, I think it is acceptable (and valid) to test based on the implementation details.

There are other tests below testing exactly that. It was more to make the test as "unit" as possible but I think it should be ok.

brettwooldridge:
Excellent.

brettwooldridge:
Looks good, thanks for the contribution!  Ready to merge?

gsmet:
Let me squash it so that we got only one commit. I'll ping you when it's done!

gsmet:
Done. You can merge it.

Thanks!

gsmet:
Hmmm, got a checkstyle issue in the new code. Fixing it.

gsmet:
OK, this time, it's ready to merge.

PR #80, opened 2014-04-30
Various assorted fixes

gsmet:
Here are various assorted fixes.

Feel free to cherry pick if not convinced by some of them.

Most of these fixes are for minor cosmetic issues or very minor performance improvements.

There is one bug though:
https://github.com/openwide-java/pgjdbc-ng/commit/dbba49971f08ca484a47172d3aebc5bcf3fcf391


kdubb:
Instead of closing in a finally, use a try-with-resources statement.

brettwooldridge:
These all look good.  In agreement with the try-with-resources.

gsmet:
> Instead of closing in a finally, use a try-with-resources statement.

I haven't done it this way as I haven't found a lot of Java7isms in the code (note that one of the finally clause was already there). But if we can do it this way, I will.

kdubb:
There are plenty of Java7isms in there; especially try-with-resources... there was some talk back when I started the project about getting it to compile with JDK6 but the reality was I wanted to target Java7 and above since I knew the adoption rate would be longer than the normal transition from 6-7.  So as it stands it's staying 7+ for the foreseeable future.

brettwooldridge:
I'm totally on board with sticking to Java 7+.

gsmet:
> Instead of closing in a finally, use a try-with-resources statement.

Done. Checkstyle OK. It should be ready to merge.

PR #82, opened 2014-05-01
OSGi bundle MANIFEST changes.

brettwooldridge:
These changes to export/import packages were necessary to get pgjdbc-ng started in my OSGi container.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,6 +50,8 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<!-- Checkstyle configuration -->
 		<linkXRef>false</linkXRef>
+		<project.osgi.exports>com.impossibl.postgres.jdbc,com.impossibl.postgres.jdbc.xa,com.impossibl.postgres.api.data,com.impossibl.postgres.api.jdbc</project.osgi.exports>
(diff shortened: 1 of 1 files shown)


brettwooldridge:
Anyone want to take a look at this before I merge it?

kdubb:
Can you explain whats going on here? lol

brettwooldridge:
In OSGi, only the packages that are "exported" with an `Export-Package` manifest header are visible to users.  Previously (last year), you merged the initial OSGi header fix, but I recently realized that 1) the headers were only added to the shaded ("complete") JAR and not the regular JAR, and 2) we have an additional package (`com.impossibl.postgres.jdbc.xa`) that needs to be visible to users now.  So, part of the fix is to added the additional package and add the `Export-Package` header to both JARs.

The `Import-Package` header is the inverse of `Export-Package`, in that unless a package is "imported" (for example, the `io.netty` packages), classes in those packages will not be loadable by the OSGi _ClassLoader_.  So without the `Import-Package` header, in OSGi the pgjdbc-ng classes throw `ClassNotFoundException` for netty classes.

brettwooldridge:
Oh, the other change was to use the `build-helper-maven-plugin`.  Because OSGi bundle (JAR) versions in the manifest header that contain a '-' character are not valid, using the `${project.version}` as the bundle version results in an invalid OSGi bundle in the case of SNAPSHOTs.  For example, `0.4.0-SNAPSHOT` is not a valid OSGi version but `0.4.0` or `0.4.0.SNAPSHOT` is.  The `build-helper-maven-plugin` creates versions components (`0`, `4`, `0`, etc.) that can be used to compose a valid bundle version header in the MANIFEST file for OSGi.

kdubb:
Ok this makes sense.  The version parsing and build helper were what was confusing me.

PR #83, opened 2014-05-01
Fix NPE in the PGPreparedStatementDelegator due to unassigned class member

brettwooldridge:
Using the XADataSource I was getting an NPE from prepared statements, this uninitialized member was the cause.


gsmet:
Hi Brett,

It's already part of my patch set: https://github.com/openwide-java/pgjdbc-ng/commit/dbba49971f08ca484a47172d3aebc5bcf3fcf391 (with the same fix for another class).

brettwooldridge:
Fixed by merge of #80, closing.

PR #84, opened 2014-05-07
Update Maven dependencies. Also update plugin dependencies.

gsmet:
Add a prerequisite for Maven version.
The output of the Maven versions plugin is more readable
with a minimal Maven version and 3.1 looks like a reasonable base
version.

Exclude the Guava classes from the checkstyle scan as they don't
respect the new checkstyle rules but we want to keep them in sync
with Guava's original classes.

PR #91, opened 2014-07-03
Rebased version of the patches of Mikko Tiihonen

gsmet:
This pull request replaces https://github.com/impossibl/pgjdbc-ng/pull/73 .

I've merged Mikko's changes in the latest devel branch.

### DIFF
--- a/src/main/java/com/impossibl/postgres/protocol/v30/ProtocolFactoryImpl.java
+++ b/src/main/java/com/impossibl/postgres/protocol/v30/ProtocolFactoryImpl.java
@@ -281,6 +281,9 @@ public class ProtocolFactoryImpl implements ProtocolFactory {
     if (e instanceof IOException) {
       io = (IOException) e;
     }
+    else if (e.getCause() == null) {
+      io = new IOException(e);
+    }
     else if (e.getCause() instanceof IOException) {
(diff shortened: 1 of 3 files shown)


kdubb:
This means I should close #73 correct?

kdubb:
Do you guys things we should add a "tag" to the Version structure so that we can keep the "devel" & "beta" tags as well?

gsmet:
Hi Kevin,

Yes, #73 is replaced by #91 + #84 (#84 is more complete and cleaner that what was proposed in the relevant part of #73: I updated all the dependencies and ignored the Guava classes in the checkstyle conf instead of reformatting them). We can close #73.

As for tagging the version, I don't see the point in it currently. The devel and beta versions of PostgreSQL are a transient process and they aren't supposed to be different from the app point of view. The only reason I can see is if someone is using the Version to display the PostgreSQL version of the server. We don't do it but it might be something someone could need.

If we do it, we could maybe name it qualifier as it's done in the Maven version convention (http://docs.codehaus.org/display/MAVEN/Versioning).

I can work on it if you think it's a good idea.

## 

Guillaume

andreak:
Is this going to be merged any time soon?

PR #93, opened 2014-08-18
Remove some synchronized stuff

gsmet:
- I removed a couple of simple StringBuilder which should have been StringBuffers
- I worked on TimestampUtils to remove the need of the synchronized keyword on a lot of methods. I think what I did is safe but it would be nice to have a second pair of eyes on it.

The only StringBuffer left in the code is used with appendReplacement and we cannot remove it.


gsmet:
@brettwooldridge could you take a look at 37721da? I know you're well versed in this sort of things.

Thanks!

PR #101, opened 2014-10-23
Fix #95 handle various orderings of IN, OUT, and INOUT parameters

### LINKED ISSUE #95: Order of parameters into a function - CallableStatement
Hello, I'm having the following experiences calling a function...
This function has all the "IN" parameters first... and the call from java works ok.. 

``` sql
CREATE OR REPLACE FUNCTION prueba.working(IN in_param_0 character varying, 
OUT out_param_1 integer, OUT out_param_2 retorno_type)
  RETURNS record AS
$BODY$

    begin
        out_param_2.codigo = '00000';
        out_param_2.mensaje = 'Este es un mensaje';
        out_param_1 = 1024;
    end;
$BODY$
  LANGUAGE plpgsql VOLATILE
  COST 100;
```

``` java
public static void working() throws SQLException, Exception {
        Connection conn = TestUtil.openDB();
        String sql = "{call working(???)}";
        CallableStatement call = conn.prepareCall(sql);
        call.setObject(1, "971238007", Types.VARCHAR);
        call.registerOutParameter(2, Types.INTEGER); 
        call.registerOutParameter(3, Types.STRUCT); 
        try {
            call.execute();
        } catch (Exception e) {
            e.printStackTrace();
            return;
        }
        System.out.println("... ... ... ... ... Llamada exitosa...");
        call.close();
        conn.close();

    }
```

My result: ... ... ... ... ... Llamada exitosa...

But, in the next example my call from java is not working... The only change is the order of the parameters...

``` sql
CREATE OR REPLACE FUNCTION prueba.not_working(OUT out_param_1 integer,
OUT out_param_2 retorno_type, IN in_param_0 character varying)
  RETURNS record AS
$BODY$

    begin
        out_param_2.codigo = '00000';
        out_param_2.mensaje = 'Este es un mensaje';
        out_param_1 = 1024;
    end;
$BODY$
  LANGUAGE plpgsql VOLATILE
  COST 100;
```

This is how I call.. 

``` java
public static void notWorking() throws SQLException, Exception {
        Connection conn = TestUtil.openDB();
        String sql = "{call not_working(???)}";
        CallableStatement call = conn.prepareCall(sql);
        call.registerOutParameter(1, Types.INTEGER);
        call.registerOutParameter(2, Types.STRUCT);
        call.setObject(3, "971238007", Types.VARCHAR);
        try {
            call.execute();
        } catch (Exception e) {
            e.printStackTrace();
            return;
        }
        System.out.println("... ... ... ... ... Llamada exitosa...");
        call.close();
        conn.close();
    }
```

I get an exception:
com.impossibl.postgres.jdbc.PGSQLSimpleException: no se pudo determinar el tipo del parámetro $1
    at com.impossibl.postgres.jdbc.ErrorUtils.makeSQLException(ErrorUtils.java:159)
    at com.impossibl.postgres.jdbc.ErrorUtils.makeSQLException(ErrorUtils.java:136)
    at com.impossibl.postgres.jdbc.PGConnectionImpl.execute(PGConnectionImpl.java:432)
    at com.impossibl.postgres.jdbc.PGPreparedStatement$1.call(PGPreparedStatement.java:183)
    at com.impossibl.postgres.jdbc.PGPreparedStatement$1.call(PGPreparedStatement.java:174)
    at com.impossibl.postgres.jdbc.PGConnectionImpl.getCachedStatement(PGConnectionImpl.java:1257)
    at com.impossibl.postgres.jdbc.PGPreparedStatement.parseIfNeeded(PGPreparedStatement.java:174)
    at com.impossibl.postgres.jdbc.PGCallableStatement.parseIfNeeded(PGCallableStatement.java:139)
    at com.impossibl.postgres.jdbc.PGPreparedStatement.execute(PGPreparedStatement.java:210)
at com.impossibl.postgres.jdbc.PGCallableStatement.execute(PGCallableStatement.java:145)
    at py.com.personal.PGDriver.ProvisionarTest.notWorking(ProvisionarTest.java:186)
    at py.com.personal.PGDriver.Main.main(Main.java:31)
I also can't call to this one that has IN, OUT, OUT, IN order... 

``` sql
CREATE OR REPLACE FUNCTION prueba.not_working_2(IN in_param_1 character varying, 
OUT out_param_1 integer, OUT out_param_2 retorno_type, IN in_param_2 character varying)
  RETURNS record AS
$BODY$

    begin
        out_param_2.codigo = '00000';
        out_param_2.mensaje = 'Este es un mensaje';
        out_param_1 = 1024;
    end;
$BODY$
  LANGUAGE plpgsql VOLATILE
  COST 100;
```

``` java
public static void notWorking2() throws SQLException, Exception {
        Connection conn = TestUtil.openDB();
        String sql = "{call not_working_2(????)}";
        CallableStatement call = conn.prepareCall(sql);
        call.setObject(1, "971238007", Types.VARCHAR);
        call.registerOutParameter(2, Types.INTEGER);
        call.registerOutParameter(3, Types.STRUCT); 
        call.setObject(4, "971238007", Types.VARCHAR);
        try {
            call.execute();
        } catch (Exception e) {
            e.printStackTrace();
            return;
        }
        System.out.println("... ... ... ... ... Llamada exitosa...");
        call.close();
        conn.close();

    }
```

com.impossibl.postgres.jdbc.PGSQLSimpleException: error de sintaxis en o cerca de «$4»
    at com.impossibl.postgres.jdbc.ErrorUtils.makeSQLException(ErrorUtils.java:159)
    at com.impossibl.postgres.jdbc.ErrorUtils.makeSQLException(ErrorUtils.java:136)
    at com.impossibl.postgres.jdbc.PGConnectionImpl.execute(PGConnectionImpl.java:432)
    at com.impossibl.postgres.jdbc.PGPreparedStatement$1.call(PGPreparedStatement.java:183)
    at com.impossibl.postgres.jdbc.PGPreparedStatement$1.call(PGPreparedStatement.java:174)
    at com.impossibl.postgres.jdbc.PGConnectionImpl.getCachedStatement(PGConnectionImpl.java:1257)
    at com.impossibl.postgres.jdbc.PGPreparedStatement.parseIfNeeded(PGPreparedStatement.java:174)
    at com.impossibl.postgres.jdbc.PGCallableStatement.parseIfNeeded(PGCallableStatement.java:139)
    at com.impossibl.postgres.jdbc.PGPreparedStatement.execute(PGPreparedStatement.java:210)
    at com.impossibl.postgres.jdbc.PGCallableStatement.execute(PGCallableStatement.java:145)
    at py.com.personal.PGDriver.ProvisionarTest.notWorking2(ProvisionarTest.java:205)
    at py.com.personal.PGDriver.Main.main(Main.java:34)

Is there any restriction about the order of the parameters? The "IN" params should always come first? Or I'm missing something?

brettwooldridge:
The existing code only works if IN or INOUT parameters occur before OUT parameters.

Given the (processed) SQL `( SELECT * FROM test_somein_someout3 ($1,$2,$3))`, where `$1` and `$2` are OUT parameters and `$3` is an IN parameter, the former could would produce processed SQL like `( SELECT * FROM test_somein_someout3 ($3))` (OUT parameters must be removed).  The problem with this is that PostgreSQL is expecting placeholders to start from `$1`.  As placeholders are processed the need to be "re-sequenced".  If `$2` is removed, the remaining placeholders should be `($1,$2)` (with `$3`→`$2`).

The existing regexes also had other issues.  For example, if there were 10 or more parameters, `SELECT ... ($1,$2,$3,$4,$5,$6,$7,$8,$9,$10)`, the regex for `$1` would also match the `$1` of `$10` and redact it, leaving `$0`.  While the new regex is quite a bit more complex, it covers every variation I would think of.

The existing regexes also had other issues, for example, when redacting `$1`, `$2`, and `$3` out of `($1,$2,$3,$4)` the result was `(,,$1)` (after the re-sequencing fix).  The regexes were adjusted to remove multiple occurrences of sequential commas, so the result is a clean `($1)`.

Additional test cases were added to cover various orderings of parameter types.  Given the complexity of the regex changes, I'm not sure how easy it is for someone to review, but if you'd like to take a look...

I can expound more on what the regex is doing if requested.

### DIFF
--- a/src/main/java/com/impossibl/postgres/jdbc/PGCallableStatement.java
+++ b/src/main/java/com/impossibl/postgres/jdbc/PGCallableStatement.java
@@ -124,17 +124,22 @@ public class PGCallableStatement extends PGPreparedStatement implements Callable
 
     sqlText = fullSqlText;
 
+    int reSeq = 1;
     for (int c = 0; c < allParameterModes.size(); ++c) {
 
       if (allParameterModes.get(c) == ParameterMode.Out) {
-        sqlText = sqlText.replaceAll("\\s*\\$" + (c + 1) + "\\s*", "");
+        sqlText = sqlText.replaceAll("\\s*\\$" + (c + 1) + "\\s*([,)])|\\s*\\$" + (c + 1) + "(\\s*)$", "$1$2");
+      }
+      else {  // re-sequence
+        sqlText = sqlText.replaceAll("\\s*\\$(" + (c + 1) + ")\\s*([,)])|\\s*\\$(" + (c + 1) + ")(\\s*)$", "\\$" + reSeq + "$2$4");
+        ++reSeq;
       }
 
     }
 
-    sqlText = sqlText.replaceAll("\\(\\s*,", "(");
-    sqlText = sqlText.replaceAll(",\\s*,", ",");
-    sqlText = sqlText.replaceAll(",\\s*\\)", ")");
+    sqlText = sqlText.replaceAll("\\(\\s*,+", "(");
+    sqlText = sqlText.replaceAll(",+\\s*,", ",");
+    sqlText = sqlText.replaceAll(",+\\s*\\)", ")");
 
     super.parseIfNeeded();
   }
--- a/src/test/java/com/impossibl/postgres/jdbc/CallableStatementTest.java
+++ b/src/test/java/com/impossibl/postgres/jdbc/CallableStatementTest.java
@@ -47,6 +47,7 @@ import java.sql.Statement;
 import java.sql.Types;
 
 import org.junit.After;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -126,6 +127,26 @@ public class CallableStatementTest {
             + "pc := pa + 1;"
             + "end;'"
             + "LANGUAGE plpgsql VOLATILE;");
+    stmt.execute("CREATE OR REPLACE FUNCTION test_somein_someout2("
+        + "pb OUT varchar,"
+        + "pc OUT int8,"
+        + "pa IN int4)"
+        + " AS "
+        + "'begin "
+        + "pb := ''out'';"
+        + "pc := pa + 1;"
+        + "end;'"
+        + "LANGUAGE plpgsql VOLATILE;");
+    stmt.execute("CREATE OR REPLACE FUNCTION test_somein_someout3("
+        + "pb OUT varchar,"
+        + "pa IN int4,"
+        + "pc OUT int8)"
+        + " AS "
+        + "'begin "
+        + "pb := ''out'';"
+        + "pc := pa + 1;"
+        + "end;'"
+        + "LANGUAGE plpgsql VOLATILE;");
     stmt.execute("CREATE OR REPLACE FUNCTION test_allinout("
             + "pa INOUT int4,"
             + "pb INOUT varchar,"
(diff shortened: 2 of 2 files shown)


kdubb:
Although I can't take the mental leap to to fully comprehend your changes to the regex's right now... the unit tests look nice.  Good work.

gsmet:
Maybe it would be even better if we compiled the regexps instead of parsing them for each call?

WDYT?

brettwooldridge:
Trouble is the regex is dynamic.  It changes every iteration of the loop.   First iteration is `\s*\$1\s*([,)])|\s*\$1(\s*)$`, next is `\s*\$2\s*([,)])|\s*\$2(\s*)$`, then `\s*\$3\s*([,)])|\s*\$3(\s*)$`.  We don't know how many replacement parameters there are.  We could create a "cache" of compiled regexes, with an integer key (1,2,3, etc.), and compile new ones "on-demand" ...

gsmet:
My remark was more for the static ones which can be easily compiled:

```
+ sqlText = sqlText.replaceAll("\\(\\s*,+", "(");
+ sqlText = sqlText.replaceAll(",+\\s*,", ",");
+ sqlText = sqlText.replaceAll(",+\\s*\\)", ")");
```

As for the cache it might be a good solution but I think there should be a hard limit where the regexps stay dynamic anyway.

Maybe it's not worth the effort and the memory though..

brettwooldridge:
I'll take a look.  Agreed on the static ones.  But even for the dynamic ones, I can envision a scenario where someone invokes a callable statement thousands of times ... and the regex compliation could impose overhead.  I also worry a bit that using regexes like this, rather than a parse-tree could lead to erroneous replacements in constants.  For example:

`SELECT * FROM table WHERE col1 < some_function(?,?) AND id='$1AS3D98YWI'`

The statement would be processed into:

`SELECT * FROM table WHERE col1 < some_function($1,$2) AND id='$1AS3D98YWI'`

and if we need to redact `$1`, the regex would also redact the `$1` out of the constant `$1AS3D98YWI`.

Without a full parse-tree of the SQL I'm not sure there is much to be done about it at this point, and it remains theoretical (but a potentially difficult bug to track down later).

brettwooldridge:
I'm going to close this, take another stab, and re-submit.

PR #108, opened 2014-10-27
Fix #96 many implicit conversions to/from String were broken.

### LINKED ISSUE #96: Using setString to set a MAC address causes IOException: invalid length
Using setString (`stmt.setString(3, "00:50:56:00:00:09");`) to set a MAC throws an IOException. From the stack trace it looks like it's been converted to bytes first. 
The setBytes method (`stmt.setBytes(3, new byte[] {0,0x50,0x56,0,0,0x09});`) works as expected.

```
java.sql.SQLException: java.io.IOException: invalid length
     at com.impossibl.postgres.jdbc.PGConnectionImpl.execute(PGConnectionImpl.java:452)
     at com.impossibl.postgres.jdbc.PGStatement.executeStatement(PGStatement.java:372)
     at com.impossibl.postgres.jdbc.PGPreparedStatement.execute(PGPreparedStatement.java:239)
     at com.impossibl.postgres.jdbc.PGPreparedStatement.executeQuery(PGPreparedStatement.java:255)
     at com.impossibl.postgres.jdbc.PGPreparedStatement.executeQuery(PGPreparedStatement.java:92)
     at com.impossibl.postgres.jdbc.PGPreparedStatementDelegator.executeQuery(PGPreparedStatementDelegator.java:134)
     at _private_
With Throwable cause: java.io.IOException: invalid length
     at com.impossibl.postgres.system.procs.MacAddrs$BinEncoder.encode(MacAddrs.java:99)
     at com.impossibl.postgres.system.procs.BinaryEncoder.length(BinaryEncoder.java:51)
     at com.impossibl.postgres.protocol.v30.ProtocolImpl.lengthOfParams(ProtocolImpl.java:723)
     at com.impossibl.postgres.protocol.v30.ProtocolImpl.writeBind(ProtocolImpl.java:507)
     at com.impossibl.postgres.protocol.v30.BindExecCommandImpl.execute(BindExecCommandImpl.java:319)
     at com.impossibl.postgres.protocol.v30.ProtocolImpl.execute(ProtocolImpl.java:369)
     at com.impossibl.postgres.jdbc.PGConnectionImpl.execute(PGConnectionImpl.java:428)
     at com.impossibl.postgres.jdbc.PGStatement.executeStatement(PGStatement.java:372)
     at com.impossibl.postgres.jdbc.PGPreparedStatement.execute(PGPreparedStatement.java:239)
     at com.impossibl.postgres.jdbc.PGPreparedStatement.executeQuery(PGPreparedStatement.java:255)
     at com.impossibl.postgres.jdbc.PGPreparedStatement.executeQuery(PGPreparedStatement.java:92)
     at com.impossibl.postgres.jdbc.PGPreparedStatementDelegator.executeQuery(PGPreparedStatementDelegator.java:134)
    at _private_
```

brettwooldridge:
Changed `SQLTypeUtils.coerceToString()` to accept a `Type` parameter and use the type-appropriate `TextCodec` if available.  Changed `SQLTypeUtils.coerce()` to handle coercions from `String` to the required type if the text parameter format is supported for that type using the type-appropriate `TextCodec`.   Changed `SQLTypeUtils.coerceToBytes()` to use type-appropriate `TextCodec` if the source value is a String and a codec exists.

Added test case to cover original issue, plus additional tests that uncovered further conversion issues.  All suite tests passing.

### DIFF
--- a/src/main/java/com/impossibl/postgres/jdbc/SQLTypeUtils.java
+++ b/src/main/java/com/impossibl/postgres/jdbc/SQLTypeUtils.java
@@ -215,7 +215,7 @@ class SQLTypeUtils {
       return coerceToBoolean(val);
     }
     else if (targetType == String.class) {
-      return coerceToString(val, connection);
+      return coerceToString(val, sourceType, connection);
     }
     else if (targetType == Date.class) {
       return coerceToDate(val, zone, connection);
@@ -270,6 +270,14 @@ class SQLTypeUtils {
     }
     else if (SQLData.class.isAssignableFrom(targetType)) {
       return coerceToCustomType(val, sourceType, targetType, typeMap, connection);
+    }
+    else if (val instanceof String && sourceType.isParameterFormatSupported(Format.Text)) {
+      try {
+        return sourceType.getCodec(Format.Text).decoder.decode(sourceType, sourceType.getLength(), null, val, connection);
+      }
+      catch (IOException e) {
+        // fall-thru
+      }
     }
 
     throw createCoercionException(val.getClass(), targetType);
@@ -517,7 +525,7 @@ class SQLTypeUtils {
     throw createCoercionException(val.getClass(), boolean.class);
   }
 
-  public static String coerceToString(Object val, Context context) throws SQLException {
+  public static String coerceToString(Object val, Type type, Context context) throws SQLException {
 
     if (val == null) {
       return null;
@@ -552,11 +560,14 @@ class SQLTypeUtils {
     else if (val instanceof Instant) {
       return ((Instant) val).disambiguate(TimeZone.getDefault()).print(context);
     }
-    else if (val instanceof byte[]) {
-      return new String((byte[]) val, context.getCharset());
-    }
     else if (val instanceof PGSQLXML) {
       return ((PGSQLXML) val).getString();
+    }
+    else if (type.isResultFormatSupported(Format.Text)) {
+      return coerceToStringFromType(val, type, context);
+    }
+    else if (val instanceof byte[]) {
+      return new String((byte[]) val, context.getCharset());
     }
     else {
       return val.toString();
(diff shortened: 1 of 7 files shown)


src/main/java/com/impossibl/postgres/jdbc/SQLTypeUtils.java
```diff
@@ -552,12 +560,15 @@ else if (val instanceof Interval) {
     else if (val instanceof Instant) {
       return ((Instant) val).disambiguate(TimeZone.getDefault()).print(context);
     }
-    else if (val instanceof byte[]) {
-      return new String((byte[]) val, context.getCharset());
-    }
     else if (val instanceof PGSQLXML) {
       return ((PGSQLXML) val).getString();
     }
+    else if (type.isResultFormatSupported(Format.Text)) {
```
kdubb, line 566: Is there a reason to put the less specific (Format support test) before the more specific (byte[] instance test)?

kdubb:
Other than the annotated question the rest seems.. :+1:

src/main/java/com/impossibl/postgres/jdbc/SQLTypeUtils.java
brettwooldridge, line 566: Yes, because the conversion of the type may be from `byte[]` to String via a specific conversion.  In the case of the `macaddr` type, `val` is a `byte[]`.  If this code runs first:

``` java
else if (val instanceof byte[]) {
   return new String((byte[]) val, context.getCharset());
}
```

we just try to convert the bytes into a string directly, but what is needed is that the `MacAddrs.TxtEncoder.encode()` should be called because it knows how to convert a byte[]-encoded MAC address into text.

kdubb, line 566: Ok I see that.  Ultimately though the fact they we are relying on a specific order seems like it might be incorrect.

kdubb, line 566: I guess my real point is that at some point, with the Type instance available, we should be able to just call "coerceToStringFromType" and ignore all else.  Maybe I am misremembering the complete details of all this.

brettwooldridge, line 566: I had the same initial thought.  At least with respect to `coerceToString()` there are only really a few special cases (I count 5), most of the existing ones could be collapsed into the final `else` (`return val.toString()`).  I haven't dug too deeply whether the special cases could be collapsed into a single case or not -- though I suspect so.

Is it the case that we can expect all types to have a TextCodec?

brettwooldridge:
I can add an enhancement for refactoring the coercion code further.

Issue #65, opened 2014-01-28
java.lang.StackOverflowError if SQL contains C-style comments


andreak:
I use 0.0.3-SNAPSHOT 5d0f7e99672c37af48ef54925fb57bc7763682fa

The C-style comment in the query below which makes it error out:

```
select coalesce(h.company_id, expense.company_id) as company_id
    , coalesce(h.companyname, expense.companyname) as companyname
    , coalesce(h.project_id, expense.project_id) as project_id
    , coalesce(h.projectname, expense.projectname) as projectname
    , coalesce(h.company_is_active, expense.company_is_active) as company_is_active
    , coalesce(h.project_is_active, expense.project_is_active) as project_is_active

/*
    , coalesce(h.company_favourite_priority, expense.company_favourite_priority) as company_favourite_priority
    , coalesce(h.project_favourite_priority, expense.project_favourite_priority) as project_favourite_priority
*/


    , h.work_time
    , h.invoiceable_time
```

java.lang.StackOverflowError
        at java.util.regex.Pattern$GroupHead.match(Pattern.java:4554)
        at java.util.regex.Pattern$Loop.match(Pattern.java:4683)
        at java.util.regex.Pattern$GroupTail.match(Pattern.java:4615)
        at java.util.regex.Pattern$BranchConn.match(Pattern.java:4466)
..
..

If I remove the comment the query works fine

brettwooldridge:
I was wondering if the regex-based parsing was fragile or not.  Time will tell just how fragile.

kdubb:
It is most likely very fragile.  I have began in implementation based on jflex/antlr but the trick is only understanding enough to parse and find actual escape sequences without trying to understand all of Postgres's syntax.  It ends up being not so easy.

brettwooldridge:
Just a thought, but you might actually find it easier to write-out the state machine by hand and hand-code the parser rather than to build a partial grammar.

kdubb:
@andreak  I added this SQL to the SQLTextTests and I cannot get it to fail as you did.  Can you make try it in that way as well?

brettwooldridge:
@andreak Can you re-test using the latest code from the `develop` branch and report whether you are still receiving this error?

andreak:
I'm still getting the error using the latest from develop

This SQL causes it, and prevents the query from reaching the server:

``` sql
SELECT
      folder.entity_id  AS folder_id
    , archive_pers.entity_id AS person_id
/*
[...]

PR #110, opened 2014-10-30
Hand written SQL-parser, based off of former regexes.

brettwooldridge:
The diffs are probably not that useful because of the size of the change.  I recommend looking at the SQLText.java file directly, it's more readable than the diff.

All tests are passing, plus an additional test from bug #65 added.

Note, this bring us back to "parsing parity" with the existing code.  However, as I was researching the PostgreSQL syntax, I noticed quite a few additions between PostgreSQL 8.x and 9.x.  Things like binary constants and unicode constants.  I do not know whether the parser handles these effective (former or current), but probably additional tests (and parsing) will need to be done in the future to cover these cases.

### DIFF
--- a/src/main/java/com/impossibl/postgres/jdbc/SQLText.java
+++ b/src/main/java/com/impossibl/postgres/jdbc/SQLText.java
@@ -58,7 +58,7 @@ public class SQLText {
     root = parse(sqlText);
   }
 
-  private SQLText(MultiStatementNode copyRoot) {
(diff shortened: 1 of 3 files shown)


gsmet:
Hi Brett,

Have you benchmarked this change?

It would be nice to check that there isn't any performance regression related to this change.

brettwooldridge:
I haven't benchmarked it, but I would expect it to perform substantially better than the regex-based parsing.

PR #114, opened 2014-11-04
Notice should be Serializable as it is used in NoticeException which is Serializable

gsmet:


jesperpedersen:

PR #115, opened 2014-11-04
Fix hardcoded usernames in tests to use the TestUtil.getUser()

gmokki:


jesperpedersen:

Issue #87, opened 2014-05-28
Connection string parse exception


slorion:
With the following connection string:

```
jdbc:pgsql://localhost:6432/test?preparedStatementCacheSize=0
```

I get: 

```
java.lang.ClassCastException: Cannot cast java.lang.String to java.lang.Integer
at java.lang.Class.cast(Class.java:3094)
    at com.impossibl.postgres.system.BasicContext.getSetting(BasicContext.java:176)
    at com.impossibl.postgres.jdbc.PGConnectionImpl.<init>(PGConnectionImpl.java:185)
    at com.impossibl.postgres.jdbc.ConnectionUtil.createConnection(ConnectionUtil.java:168)
    at com.impossibl.postgres.jdbc.PGDriver.connect(PGDriver.java:60)
    at com.impossibl.postgres.jdbc.PGDriver.connect(PGDriver.java:50)
    at java.sql.DriverManager.getConnection(DriverManager.java:571)
    at java.sql.DriverManager.getConnection(DriverManager.java:215)
    at org.sql2o.GenericDatasource.getConnection(GenericDatasource.java:48)
    at org.sql2o.Connection.createConnection(Connection.java:240)
```

Context: I am trying to use this driver with pgbouncer with pool_mode=statement

slorion:
Thank you for working on a fix right away! In the meantime, I am using AbstractDataSource.setPreparedStatementCacheSize, but, unless mistaken, it does not seem to have the effect I was hoping for, namely what is described here: http://engineering.clever-cloud.com/java/2013/05/27/pgjdbc-now-pgbouncer-compliant.html. If I am using the wrong setting, would you mind telling me the right one, if one exists ?

brettwooldridge:
You are correct, the prepared statement cache size setting will not have the effect you are seeking. We'll need to investigate the semantics of the changes that were made to the Postgresql JDBC driver.

slorion:
Ok, thanks again. The app is still in dev, so I can use pool_mode=session for now.

slorion:
Hello Brett,

Any news regarding support for preparedStatementCacheSize=0 ? That missing feature basically prevents this library from being used with pgBouncer, because being required to use pool_mode=session defeats one of the main advantage of using pgBouncer. If it is something non trivial to implement, I could give a hand.

kdubb:
If you are not already the preparedStatementCacheSize=0 request should be formalized into an enhancement for disabling the cache

[...]

PR #119, opened 2014-11-13
Fix URL parameter parsing when an int/Integer is required.

gsmet:
Based on a patch by Brett.

see https://github.com/impossibl/pgjdbc-ng/issues/87


jesperpedersen:

Issue #148, opened 2015-04-06
setAutoCommit(true) + setNetworkTimeout(n) causes hang on PostgreSQL shutdown


csdillard:
When network timeout is set to a positive value and auto commit is `false`, the network timeout never seems to get triggered.

A simple test case goes like this:  Make a new `PGDataSource`.  Set its timeout with `setNetworkTimeout(1000)` (for example).  Get a `Connection` from it and call its `setAutoCommit(false)`.  Then loop forever, doing `SELECT 1`, `commit()`, then sleep.  While it's looping, shut down PostgreSQL.

Expected behavior is that shortly after PostgreSQL shutdown, an exception would be thrown.  But it just hangs.

If you comment out the `setAutoCommit(false)` (and the `commit()`), then it works as expected.

This was tested against the 0.5-SNAPSHOT version that's current as of 2015 Apr 06 (at 12pm Central).  Full code of my test case:

``` java
package test;

import java.sql.Connection;
import java.sql.PreparedStatement;
import java.sql.ResultSet;

import com.impossibl.postgres.jdbc.PGDataSource;

public class TimeoutFailureTest {
    public static void main(String[] args) throws Throwable {
        boolean BREAK_IT = true;
        PGDataSource dataSource = new PGDataSource();
        dataSource.setHost("localhost");
        dataSource.setDatabase("tester");
        dataSource.setUser("tester");
        dataSource.setPassword("password");
        dataSource.setPort(5432);
        dataSource.setApplicationName("Broken Connection Test");
        dataSource.setNetworkTimeout(1000);

        Connection connection = dataSource.getConnection();
        if (BREAK_IT)
            connection.setAutoCommit(false);

        while (true) {
            try (PreparedStatement ps = connection.prepareStatement("SELECT 1");
                    ResultSet rs = ps.executeQuery()) {
                rs.next();
            }
            if (BREAK_IT)
                connection.commit();
            System.out.println("query completed.  shut down postgresql now...");
            Thread.sleep(1000);
        }
    }
}
```

jesperpedersen:
I have added a test case: https://github.com/impossibl/pgjdbc-ng/commit/9a7341a2ee8f6cc1b07fd7b2347fdfbae74c7eee that you can modify

csdillard:
Thanks for adding that test case.  Unfortunately (for me), I can't make it break!

[...]

PR #152, opened 2015-04-07
ConnectionTest.java edits to show network timeout failure on terminated backend

csdillard:
The edited test case now demonstrates [the bug here](https://github.com/impossibl/pgjdbc-ng/issues/148), namely that shutting down a server backend will keep network timeout from working as expected.

Though to be honest, it's not clear to me whether this is a bug in the operating system, a bug in Java's NIO, a bug in the driver, or just a general problem with TCP networking.


jesperpedersen:
That is a bit different, since the connection will be dead, and hence a SQLException will be thrown.

I have made some changes though to detect that, and changed the test case.

Issue #127, opened 2015-01-10
Allow disabling prepared statement cache with prepareThreshold=0


slorion:
Please consider making it possible to disable the prepared statement cache via the connection string parameter `?prepareThreshold=0`. This would allow pgjdbc-ng to work with PgBouncer in transaction pooling mode as explained here: http://engineering.clever-cloud.com/java/2013/05/27/pgjdbc-now-pgbouncer-compliant.html. Some more background: http://www.depesz.com/2012/12/02/what-is-the-point-of-bouncing/

jesperpedersen:
Supported in version 0.5

PR #155, opened 2015-04-09
Fix prepared statement cache disabled scenario

jesperpedersen:


slorion:
I just saw this commit. Does it fix https://github.com/impossibl/pgjdbc-ng/issues/127 ?

jesperpedersen:
@slorion Yes, it should - we will verify all existing issue reports once the 0.5 release is out

PR #157, opened 2015-04-10
Performance improvement

jesperpedersen:


src/main/java/com/impossibl/postgres/jdbc/PGConnectionImpl.java
```diff
@@ -341,18 +341,21 @@ String getNextPortalName() {
...
       WeakReference<PGStatement> statementRef = statementRefIter.next();
-
-      if (statementRef.get() == null || statementRef.get() == statement) {
+      PGStatement s = statementRef.get();
+      if (s == null) {
         statementRefIter.remove();
       }
+      else if (s == statement) {
+        statementRefIter.remove();
+        found = true;
```
brettwooldridge, line 356: Why not `break` here, and do away with the `boolean` entirely?

jesperpedersen, line 356: Thanks - merged

src/main/java/com/impossibl/postgres/jdbc/PGResultSet.java
```diff
@@ -2055,9 +2061,15 @@ int move(String type, Object loc) throws SQLException {
 
     result = null;
 
-    String sql = String.format("MOVE %s %s IN %s", type, loc, cursorName);
+    StringBuilder sb = new StringBuilder();
```
brettwooldridge, line 2064: If this change is primarily about _performance_, the default `StringBuilder()` constructor initializes with an initial capacity of 16 characters.  The string here is basically 16 characters _before_ the parameters are inserted.  We don't want to waste space, but at the same time we don't want to immediately have to expand the capacity either.  `StringBuilder` will double the capacity when it grows, so it is probably safe to start out with `new StringBuilder(32)` at a minimum, or probably safer, `new StringBuilder(48)`.

src/main/java/com/impossibl/postgres/jdbc/PGStatement.java
```diff
@@ -249,23 +249,25 @@ void closeResultSets() throws SQLException {
...
       WeakReference<PGResultSet> resultSetRef = resultSetRefIter.next();
-
-      if (resultSetRef.get() == null || resultSetRef.get() == resultSet) {
+      PGResultSet rs = resultSetRef.get();
+      if (rs == null) {
+        resultSetRefIter.remove();
+      }
+      else if (rs == resultSet) {
         resultSetRefIter.remove();
+        found = true;
```
brettwooldridge, line 263: Same comment as above.  Why not `break`?

src/main/java/com/impossibl/postgres/jdbc/PGResultSet.java
jesperpedersen, line 2064: I'll take those in a future PR - first concern was to get rid of String.format() which is dead slow, and hence shows up in performance profiles

PR #165, opened 2015-05-08
Move project to Java 8, and add JDBC 4.2 stubs

jesperpedersen:
Subject pretty much says it all - all JDBC 4.2 stubs are added with NOT_IMPLEMENTED for now, except Driver.deregister() which calls our cleanup() method.

@kdubb: Yes/No ?
@brettwooldridge: Yes/No ?
Community: Yes/No ?

I vote yes, as Java 7 is EOL, and we have not yet reached 1.0. Alternative, we can branch, but that is double maintenance IMHO.

Vote now :)


kdubb:
What about using the Maven preprocessor and building a Java 7 and Java 8 version?

jesperpedersen:
Yeah, another option. But most people will likely move to Java 8 due to 7 being EOL.

harbulot:
> Yeah, another option. But most people will likely move to Java 8 due to 7 being EOL.

I'm not sure people move so fast. (There are still places running Java 6, unfortunately.)

In addition, please note that [Java 7 is only in "End of Public Support" mode, not EOL](https://spring.io/blog/2015/04/01/ongoing-support-for-java-7-and-even-java-6).

Another consideration is that OpenJDK 8 isn't available in the current [Ubuntu LTS realease](https://wiki.ubuntu.com/LTS) (14.04). The next LTS release will probably be in April 2016, but even then, many people will probably still be using the 12.04 or 14.04 LTS releases.

kdubb:
The reason I'm "for" the complexities of a preprocessor and producing both 7 & 8 versions is simply because even if we missed the window this time.  Moving to 9, 10, etc will still require some deft maneuvering in the future.  I'd like to figure out that solution.

brettwooldridge:
I would suggest compiling under Java 8 with a compile target of Java 7.  Java 6 users can stick to the PostgreSQL provided driver.

This approach allows implementation of Java 8 JDBC APIs while still running on Java 7.

kdubb:
@brettwooldridge Does that "just work" because Java only ever adds methods to the javax.sql interfaces?

brettwooldridge:
Yes, it just works.  For the reasons you stated above.

davecramer:
Interesting, so the only issue is when the version number of the bytecode
changes ?

Dave Cramer

kdubb:
@davecramer  I remembered reading about this from some old Eclipse documentation "evolving Java APIs" or something like that.  It seems if this is the case most of the complexity of the current drivers setup may be unnecessary?

davecramer:
@kevinwooten apparently so :( There is just the issue of changing the
bytecode versions. But that has to be easier than what we do now

Dave Cramer

kdubb:
@davecramer Yes sir.  After I thought about it a bit I think what @brettwooldridge said is accurate; not that I'm necessarily doubting him.  From a compiling perspective it seems to not work.  Although, compiling under Java 8 targeting 7 means that all methods would be properly satisfied when running under Java 7.  It would also have some extra (essentially private or unknown) methods that fulfill the Java 8 contract.

kdubb:
@jesperpedersen I think if you change the pom.xml to target 1.7 and require 1.8 this PR should essentially be good to go.

brettwooldridge:
@jesperpedersen @kdubb @davecramer One of the projects I manage, [HikariCP](https://github.com/brettwooldridge/HikariCP), is a JDBC connection pool so it naturally needs to implement the `Connection`, `Statement`, etc. interfaces.

We compile everything under Java 8, with a target of `1.6` for a "HikariCP-java6" artifact (also used on Java 7), and with a target of `1.8` for a Java 8 artifact ("HikariCP" proper).

As you noted, the classes end up with "additional methods" under Java 7, but these appear as "unadvertised" public methods that are simply inaccessible when invoking through the `javax.sql` interfaces.

The only thing that the runtime cares about is that the class version is 51 (on Java 7), which is what the compile target gives you.

EDIT: note that the interesting thing is that even with a compile target of Java 7, but running under Java 8, because the interfaces come from the runtime itself, all Java 8 methods are accessible/invocable.

kdubb:
@brettwooldridge Great news. Simplifies so much.

jesperpedersen:
@kdubb What "require" ? `<source>1.8</>` requires a ``<target>1.8</>` and setting both to 1.7 will only work with Java 8, as java.sql.SQLType is new, so Java 7 compile will fail

harbulot:
`<source>1.8</source>` and `<target>1.7</target>` will indeed fail.

I'm able to get this code to compile and produce bytecode with class version 51 (Java 7) by using `<source>1.7</source>` and `<target>1.7</target>` in conjunction with a Java 8 `JAVA_HOME` when launching Maven. This presumably puts the Java 8 `rt.jar` on the classpath so as to satisfy the `java.sql.SQLType` dependency and so on. (Compilation with a Java 7 `JAVA_HOME` fails.)

I guess the side effect of this would be an element of confusion for people getting this code and trying to compile it with Java 7 (since it would have a source/target of 1.7).

jesperpedersen:
There is no rush on this PR, since all the JDBC 4.2 methods are not implemented.

Steps forward:
1) Keep using Java 7 until one or more JDBC 4.2 methods are implemented. Then switch to Java 8
2) Switch to Java 8
3) Do a dual compile project, where Java 7 and Java 8 are detected and the correct classes compiled

This PR can be used as a base for any of those.

kdubb:
My thought is nix the idea of "Java 7 or 8 detected" and once we start using this PR, and thus requiring Java 8, just notate it properly in the README.  If it fails to compile for somebody under Java 7 they can read and see it requires them to install Java 8 for compilation only.  Just simplifies so much.

jesperpedersen:
@kdubb Agreed, but lets wait until after 0.6

He-Pin:
@jesperpedersen the 0.6 version have been released now,so how about this?

kdubb:
@brettwooldridge I'm about to start fooling around with implementing this.  As @vlsi pointed pointed out, the java.sql.SQLType is a new interface in Java 8. How would the this class survive loading in a Java 7 environment?

kdubb:
@jesperpedersen @brettwooldridge I've been playing around with Java 8 support using a preprocessor. It dawned on me that since Java 7 is EOL (correct?)... maybe we just jump to supporting Java 8 only. Thoughts?

brettwooldridge:
I would say drop Java 7.  I'm getting ready to drop Java 7 support in HikariCP as well.

jesperpedersen:
Maybe cut one more release with Java 7 support, since there has been a lot of improvements over 0.6.

But I'm ok with going straight to Java 8 too, as it is simpler to code/release that way.

jorsol:
Drop Java 7 support, HikariCP has already do it.

PR #167, opened 2015-05-29
Strict mode

jesperpedersen:
I have been looking at how to deal with the

```
(byte[])rs.getObject(1);
```

based test cases in PreparedStatementTestCase without messing everything up, and maintaining good defaults. I wonder why I added those ;)

Added a "strict mode" - default is false - to force all InputStream's to be converted to byte[] for the getObject(int) calls. Other changes maybe added for this mode in the future.

@kdubb WDYT ?

### DIFF
--- a/src/main/java/com/impossibl/postgres/jdbc/PGResultSet.java
+++ b/src/main/java/com/impossibl/postgres/jdbc/PGResultSet.java
@@ -874,6 +874,9 @@ class PGResultSet implements ResultSet {
 
     Class<?> targetType = mapGetType(type, map, statement.connection);
 
+    if (statement.connection.isStrictMode() && InputStream.class.equals(targetType))
+      targetType = byte[].class;
+
     return coerce(get(columnIndex), type, targetType, map, statement.connection);
   }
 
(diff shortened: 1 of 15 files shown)


src/main/java/com/impossibl/postgres/jdbc/PGResultSet.java
```diff
@@ -874,6 +874,9 @@ public Object getObject(int columnIndex, Map<String, Class<?>> map) throws SQLEx
 
     Class<?> targetType = mapGetType(type, map, statement.connection);
 
+    if (statement.connection.isStrictMode() && InputStream.class.equals(targetType))
```
kdubb, line 877: At the very least should this not be in mapGetType?

I might advocate for a bolder change.  Does the spec say, or is it reasonable to assume that, that getObject for BINARY/VARBINARY must be returning byte[]? If so, I would map type to it's java SQL type (via SQLTypeMetadata) and translate it this way in all cases and forego the "strict" mode.

jesperpedersen, line 877: It assumes that BINARY and VARBINARY maps to two different database types -- we use "bytea" for both of course.

Ideally getObject(int) shouldn't be used for byte/input streams, but is that really tested ;) So, it is really database + JDBC driver specific what is done.

The reason I did it this way is that assuming InputStream would hit more use-cases for getObject(int) than byte[]. But we need the latter in some way.

Of course this needs documentation once we decide how to get those 4 test cases passing

PR #170, opened 2015-05-31
Please add support for SSLParameters

harbulot:
Here is a patch that allows connections to set an [`SSLParameters`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLParameters.html) instance, which can be used to choose the SSL/TLS protocol and the cipher suites (amongst other settings).

``` java

Class.forName("com.impossibl.postgres.jdbc.PGDriver");
String url = "jdbc:pgsql://.../...?ssl.mode=VerifyFull";
Properties props = new Properties();
SSLParameters sslParameters = new SSLParameters();
sslParameters.setProtocols(new String[] { "TLSv1.1" });
//sslParameters.setCipherSuites(new String[] { ... });
props.setProperty("user", "...");
props.setProperty("password", "...");
props.put("sslParameters", sslParameters);
Connection conn = DriverManager.getConnection(url, props);
Statement stmt = conn.createStatement();
// The sslinfo extension (from contrib) need to be install on the DB for these functions to work.
ResultSet rs = stmt.executeQuery("SELECT ssl_version() || ' - ' || ssl_cipher()");
if (rs.next()) {
    System.out.println("SSL/TLS version and ciphers: "
            + rs.getString(1));
}
conn.close();
```

I've also renamed the `SSL_PROTOCOL` constant into `SSLCONTEXT_PROTOCOL` in the `SSLEngineFactory`. The [`SSLContext` protocol isn't necessarily the protocol that's going to be used for the connection](http://docs.oracle.com/javase/8/docs/technotes/guides/security/SunProviders.html#SunJSSE_Protocols).

Since Java 7, they can also be used to implement host name verification directly (at least according the LDAPS and HTTPS conventions). Of course, that would be redundant with the existing host name verification mechanism, but it also checks for names in the subject alternative names. There is currently no support for the more generic [RFC 6125](https://tools.ietf.org/html/rfc6125) here in Java, but "HTTPS" is probably close enough, and could also generally be applied to PostgreSQL connections: `sslParameters.setEndpointIdentificationAlgorithm("HTTPS")`.

### DIFF
--- a/src/main/java/com/impossibl/postgres/protocol/ssl/SSLEngineFactory.java
+++ b/src/main/java/com/impossibl/postgres/protocol/ssl/SSLEngineFactory.java
@@ -67,9 +67,9 @@ import javax.security.auth.callback.CallbackHandler;
 public class SSLEngineFactory {
 
 
-  private static final String TRUST_MANAGER_FACTORY_TYPE = "PKIX";
-  private static final String SSL_PROTOCOL = "TLS";
-  private static final String KEY_STORE_TYPE = "JKS";
+  private static final String TRUST_MANAGER_FACTORY_TYPE = TrustManagerFactory.getDefaultAlgorithm();
+  private static final String SSLCONTEXT_PROTOCOL = "TLS";
+  private static final String KEY_STORE_TYPE = KeyStore.getDefaultType();
   private static final String SSL_DIR_NAME = "postgresql";
   private static final String CERTIFICATE_FACTORY_TYPE = "X.509";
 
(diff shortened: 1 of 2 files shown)


jesperpedersen:
Thanks for working on this !

I would like to see a consolidation of the SSL functionality, where each option is documented, and clear. Currently there is no SSL support in AbstractDataSource either, so that needs to be exposed.

Do you have an interest in attacking this ?

jesperpedersen:
After having looked at this, I think the best solution is to expose the needed configuration parameters as we do today, and transform them internally to SSLParameters.

However, I would like to see some community feedback on the current setup before we add more settings. If you have any concrete parameter that you want to have exposed I think the best way forward is to file an issue on it first for discussion.

We should also switch to use Netty's SSL setup in `SSLEngineFactory` in order to get support for OpenSSL.

As such, I'll close your PRs for now for further discussion on the created issues.

Thanks for your contribution on this.

PR #171, opened 2015-05-31
Use JVM default trust store by default

harbulot:
This is a further development of pull request #170.

Instead of trying to load a `root.crt` file by default (which would be very specific to PostgreSQL), this would use the JVM default trust store when no CA bundle file is specified. In a Java context, this probably makes more sense.

(Similarly, it would also make sense to be able to pass an `SSLContext` parameter to initialise a possible client certificate, since loading private keys and client certificates from PEM files is less common from a Java point of view. This could be a further development, but that might change the default values some people currently use.)


jesperpedersen:
Thanks for working on this !

See my comment on #170 for more comments.

jesperpedersen:
See my comments on #170

PR #174, opened 2015-06-08
Multiple platform and locale related test errors running on Windows 7

frode-carlsen:
Fix for eol error in Blobtest
Fix for decimal separator locale dependency in codectest (decimal separator is always parsed as dot '.' when parsing string, but may have been rendered with comma or other depending on locale.  
Fix for race condition in connectiontest.testKillConnection due to pg_terminate_backend may not complete synchronously.

### DIFF
--- a/src/main/java/com/impossibl/postgres/api/data/Interval.java
+++ b/src/main/java/com/impossibl/postgres/api/data/Interval.java
@@ -37,7 +37,6 @@ import java.util.StringTokenizer;
 import static java.lang.Double.parseDouble;
 import static java.lang.Integer.parseInt;
 import static java.lang.Math.round;
-import static java.lang.String.format;
 import static java.util.concurrent.TimeUnit.HOURS;
 import static java.util.concurrent.TimeUnit.MICROSECONDS;
 import static java.util.concurrent.TimeUnit.MINUTES;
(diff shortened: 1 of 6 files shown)


jesperpedersen:
LGTM.

@kdubb Ok, to merge ?

kdubb:
Seems like the period or comma should be driven by the Postgres context. This is forcing it to a period regardless of current locale instead correct?

jesperpedersen:
@kdubb Yes, it does. It depends on what the text representation of Interval should be. Forcing it to stable format across platforms seems like a better option to me.

kdubb:
@jesperpedersen Doesn't the Postgres server settings drive the choice though?

jesperpedersen:
@kdubb http://www.postgresql.org/docs/9.4/static/functions-datetime.html uses '.' in its examples

jesperpedersen:
@kdubb But true, maybe there is something more to it...

kdubb:
http://www.postgresql.org/message-id/4D2422BE.1060901@2ndQuadrant.com

Follow that rabbit hole and it seems that in 2011 PG only supported and output a period regardless of locale setting.  If that's still the case this is the correct fix.

jesperpedersen:
@kdubb It doesn't eat a ',' for sure in any lc_numeric :)

jesperpedersen:
Merged, thanks for the contribution !

frode-carlsen:
yw, thx

PR #177, opened 2015-06-10
Allow Binary and Text Encoder/Decoders to be extended

frode-carlsen:
Changed visibility on BinaryEncoder++ to protected from package visible so it's possible to reuse these in other projects needing to extend codecs.

Added generating source jars to maven pom.xml to make it easier to use as dependencies in development


frode-carlsen:
Hi,
Can you please merge this PR?  It includes changes to simplify implementation of postgis datatypes library, and avoid excessive code duplication.

jesperpedersen:
If you do development on a feature it is easier to eliminate merge commit, which should be removed by a rebase before a pull request. Also squash similar commits

frode-carlsen:
will do

PR #181, opened 2015-06-16
minimal support for external extensions to pgjdbc datatypes

frode-carlsen:
Since #178 is on hold, I'd like to make the following minor changes to the existing project to ease creating external datatype extensions to the library.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,7 +84,21 @@
 	</configuration>
       </plugin>
       <plugin>
-	<groupId>org.codehaus.mojo</groupId>
+    <groupId>org.apache.maven.plugins</groupId>
+    <artifactId>maven-source-plugin</artifactId>
+    <version>2.4</version>
+    <executions>
+      <execution>
+        <id>attach-sources</id>
+        <goals>
+          <goal>jar</goal>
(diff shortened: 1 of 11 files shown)


kdubb:
Does this conform to our checkstyle?  I thought "protected" should be on the same line as the method definition.

frode-carlsen:
interesting, it passed the checkstyle tests so looks like that's not enforced. 
I used automated refactoring to make the change, so I have no qualms about changing it to be on the same line.  Will check in

jesperpedersen:
@frode-carlsen Could you fix the indentation of the pom.xml file change ? Then rebase, and squash. Thanks

frode-carlsen:
ok, made some cosmetic changes to make it align same as plugins with above/below, but the whole pom.xml needs reformatting.  It's a mix of tabs and spaces and improperly aligned child elements.  Can do it here, but think perhaps it should be a different PR?

jesperpedersen:
@frode-carlsen Just format to spaces here, but squash your commits too

frode-carlsen:
how about this?

jesperpedersen:
Rebased, and merged. Thanks for your contribution !

PR #208, opened 2015-10-16
SSL for datasources

jesperpedersen:
Here is the initial SSL support for data sources.

It follows the existing functionality for the connection parameters, which can be documented in README.md now.

I'll likely do further changes, but SSLDataSourceTest is passing - although it is a bit hacky in its test setup.

Ideally this will go in after the 0.6 release, but comments are welcome now


jesperpedersen:
Merged.

Possible enhancements can follow as separate pull requests

PR #213, opened 2015-12-09
junit tests for package com.impossibl.postgres.system.tables

dstipp:
As well as additional coverage around Versions

I'm not sure how particularly useful these will be, but I didn't see any coverage here and I want to change some things around here...

### DIFF
--- a/src/test/java/com/impossibl/postgres/system/tables/PgAttributeTest.java
+++ b/src/test/java/com/impossibl/postgres/system/tables/PgAttributeTest.java
@@ -1,0 +1,175 @@
+/**
+ * Copyright (c) 2013, impossibl.com
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *  * Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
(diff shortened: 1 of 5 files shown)


src/test/java/com/impossibl/postgres/system/tables/PgAttributeTest.java
```diff
@@ -0,0 +1,175 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Created by dstipp on 12/8/15.
+ */
+@RunWith(JUnit4.class)
+public class PgAttributeTest {
```
kdubb, line 49: @dstipp  Is there a reason you specifically marked this JUnit4? Our dependencies take care of that.

kdubb, line 49: Hahah never mind.  All of our tests specify this.  Sometimes when you haven't looked at something for a while... well you know.  Needless to say these are unnecessary and someday could be removed.

dstipp, line 49: Force of habit at this point. :-)

PR #214, opened 2015-12-09
Add Travis CI support.

dstipp:
I added Travis CI to my local copy to aid in my local development.  Might be useful here as well.

(Not done in container format on purpose so I can support other versions besides the subset that Travis CI offers in addons.)

### DIFF
--- a/.travis/travis-install-dependencies
+++ b/.travis/travis-install-dependencies
@@ -1,0 +1,38 @@
+#!/bin/sh
+# Modified version of
+# https://github.com/ckan/ckan/blob/master/bin/travis-install-dependencies
+
+set -e
+set -x
+
+# Drop Travis' postgres cluster if we're building using a different pg version
+TRAVIS_PGVERSION='9.1'
+
+if [ ${PGVERSION} != ${TRAVIS_PGVERSION} ]
+then
+  sudo -u postgres pg_dropcluster --stop ${TRAVIS_PGVERSION} main
+fi
+
+# Install PostgreSQL from apt.postgresql.org
+sudo apt-get update -qq
+sudo apt-get install postgresql-${PGVERSION} postgresql-contrib-${PGVERSION}
+
+if [ ${PGVERSION} = '8.4' ]
+then
+  # force postgres to use 5432 as it's port
+  sudo sed -i -e 's/port = 5433/port = 5432/g' /etc/postgresql/8.4/main/postgresql.conf
+fi
+
+sudo service postgresql restart ${PGVERSION}
+
+sudo tail /var/log/postgresql/postgresql-${PGVERSION}-main.log
+
+# Setup postgres' users and databases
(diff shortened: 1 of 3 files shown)


kdubb:
I was just fooling around with this... awesome!  I'll look into in a bit.

.travis/travis-install-dependencies
```diff
@@ -0,0 +1,38 @@
...
+if [ ${PGVERSION} != ${TRAVIS_PGVERSION} ]
+then
+  sudo -u postgres pg_dropcluster --stop ${TRAVIS_PGVERSION} main
+fi
+
+# Install PostgreSQL from apt.postgresql.org
+sudo apt-get update -qq
+sudo apt-get install postgresql-${PGVERSION} postgresql-contrib-${PGVERSION}
+
+if [ ${PGVERSION} = '8.4' ]
```
kdubb, line 20: We don't support less than 9.1... what's this for?

dstipp, line 20: We're exploring adding support for that version and contributing it back.  (Hurray for legacy applications...)

kdubb, line 20: Yikes... that's going to be a bit of work.

Side note: I have a half written response to the email you sent me about this.  I will finish and respond soon.

kdubb:
@dstipp  You say not if container format.  I'm a Travis newb and I see we get a warning to use that format.  What exactly is the issue with using containers?

dstipp:
In travis parlance, if you use container format, you agree that you can't make changes to the system (no sudo, etc.) in exchange for caching and faster CI startup times.  It does, however, forces you to use only specific versions of software they specify.  The reason I did this was for future flexibility. (Though it looks like the Travis 'apt' plugin helps get around a lot of the reason that I was using 'sudo'.) Another side effect, though, is that if one is in container format, I don't think you can edit the configuration of Postgres which might make some things more difficult.

You shouldn't be getting a warning (I certainly don't see one on my branches...)

If you'd prefer, I can submit another PR that makes this just use container formats until 9.0/8.4 support would be pushed.

kdubb:
Not a "warning" but a notice when I view the logs on Travis about running on "older" format.  I didn't know the differences.

I'm not sure what I prefer.  I will say that we shouldn't need any "setup" beyond creating a database and installing store and pgcrypto (if available).  So that might make the decision easier.

Right now I am attempting to get Travis to deploy snapshots to maven but it's slow going as I am no Travis expert at all.

dstipp:
Oh, yea, that banner.  Hrm.  Sudo's only needed to reconfigure postgres itself (pg_hba, etc.).  Extension loading just works.  Let me dig into the 9.1 test / killing issue and see how that needs to be resolved.

As for Travis auto-deploying snapshots to Maven - it can be done, but is complicated to do right.  This might be useful reading on the matter:  https://github.com/dropwizard/dropwizard/issues/440

PR #223, opened 2016-01-11
Fixing NPE when BlobOutputStream#close() called more than once

lorenzodee:
I encounter a `NullPointerException` when calling `close()` more than once on the output stream returned via java.sql.Blob#setBinaryStream(long)  (created from Connection#createBlob()). Since I wrap the output stream inside a buffered output stream, I end up calling `close()` more than once. This also happens when using Java 7's `try`-with-resources syntax. Please see tests (added in [BlobTest.java](src/test/java/com/impossibl/postgres/jdbc/BlobTest.java)) that mimics my code.

### DIFF
--- a/src/main/java/com/impossibl/postgres/jdbc/BlobOutputStream.java
+++ b/src/main/java/com/impossibl/postgres/jdbc/BlobOutputStream.java
@@ -90,6 +90,8 @@ public class BlobOutputStream extends OutputStream {
 
   @Override
   public void close() throws IOException {
+    if (lo == null) return;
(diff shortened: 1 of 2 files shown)


jesperpedersen:
Thanks for your contribution !

Could you expand the `lo == null` checks in BlobOutputStream, since trying to write and flush after a close should result in an IOException ?

kdubb:
Haven't heard from @lorenzodee after @jesperpedersen requested changes.  So I added them and made PR #258

PR #227, opened 2016-01-15
Add ability to convert between arrays and SQLData structs

gregb:
This change enables:
1. Postgres arrays to be converted to and from SQLData implementing objects using the ResultSet.getObject(x, Class<? extends SQLData>) method.
2. Geometry items which the driver already parses into a double[] to be returned (and passed as parameters) as custom objects also implementing SQLData.

My use case was actually item 2, where item 1 was just a bonus prerequisite.  Examples of both in unit tests.

Thanks,
--Greg


jesperpedersen:
Thanks for your contribution !

Could you clear up the checkstyle issues (f.ex `mvn clean package`) ? And squash the commits, and push -f. Thanks

kdubb:
One other small thing is... the driver code doesn't use "final" argument modifiers in general.  I know there are debates about wether they are useful or not in the modern java compiler but since the precedent is that we don't use them, I'd like to keep it consistent until we do decide to change (unless specifically needed for a language feature).

So, in other words, please remove the "final"s before all the arguments. Also, thanks for your contribution!

gregb:
First, apologies, I ran the unit tests straight from eclipse and did not attempt a maven package.  Secondly, I'm not that familiar with checkstyle -- is there any way to apply your formatting preferences [checkstyle, eclipse, or otherwise] automatically?  They're a bit non-standard, and I'm not sure I'm going to find the time to go though so many lines and guess and recheck what it wants.

Thanks!

kdubb:
I began pulling your original commit in an attempt to clean it up and merge it.  Then I realized I misunderstood what you were enabling.

With your code you are passing an `SQLData` derived class to a database parameter typed as a `text[]`. This is a bit too "automatic" and opens up a bunch of questions about how people should be using `SQLData`. The correct way to use it would be to create a `person_name` type in the PostgreSQL server and have an equivalent `PersonName` in Java code; not to try and auto convert a `PersonName` into a `text[]`.

PR #233, opened 2016-02-09
Use strong references for notification listeners

kdubb:
These objects do not need to be tracked by a housekeeper therefore do not need the be referenced weakly.


jesperpedersen:
I think we should clear out `notificationListeners` as part of PGConnectionImpl.close()

jesperpedersen:
@kdubb Do you want me to finish this up ?

aergul:
We were seeing notifications stop after a few minutes with 0.6 release and I can confirm that snapshot with #233 has been running for over 14 hours now without any notifications lost. Thanks!

Issue #252, opened 2016-05-10
getting npe while executing prepared statement 


abhishekparwal:
```
java.lang.NullPointerException
    at com.impossibl.postgres.jdbc.PGPreparedStatement.executeBatch(PGPreparedStatement.java:378)

```

which points to 

`if (!allowBatchSelects() && resultBatch.command.equals("SELECT")) {
`
using 0.6 version of pgjdbc-ng driver

getting npe while executing prepared statement 

```
insert 
    into
        person
        (id, role) 
    values
        (?, ?)

```

rjohnsondev:
I'm running into this pretty frequently myself, seems pretty random and no idea how to reliably replicate at this point.

kdubb:
Clearly code that reproduces it would be grand.  I'd settle for a stack trace.

rjohnsondev:
Yeah, will continue to dig; I was getting it pretty consistently during a 2-hour import process, and managed to confirm that it's the `command` which is null. The last import worked to completion however and I don't have a full stack trace on hand; the line pasted by abhishekparwal was the only line in the driver itself however. I'm now running the latest code from develop, and have added a couple of additional checks to try and get a better idea of the state of things when that null does creep in.  Will let you know once it happens again and I have more info.

rjohnsondev:
Okay, managed to trigger one after a couple of hours. Here is a full stack trace; this was caught by the addition of:

``` java
          if (resultBatch.command == null) {
            System.out.println("Got a null batch command!");
            Thread.dumpStack();
          }
```

in `PGPreparedStatement.java:377`.

Apologies for the clojure noise:

``` java
Got a null batch command!
java.lang.Exception: Stack trace
    at java.lang.Thread.dumpStack(Thread.java:1329)
    at com.impossibl.postgres.jdbc.PGPreparedStatement.executeBatch(PGPreparedStatement.java:384)
    at clojure.java.jdbc$execute_batch.invokeStatic(jdbc.clj:426)
    at clojure.java.jdbc$execute_batch.invoke(jdbc.clj:419)
    at clojure.java.jdbc$db_do_execute_prepared_statement$fn__769.invoke(jdbc.clj:792)
    at clojure.java.jdbc$db_transaction_STAR_.invokeStatic(jdbc.clj:637)
    at clojure.java.jdbc$db_transaction_STAR_.invoke(jdbc.clj:585)
[...]

PR #270, opened 2016-07-13
Fix race condition for no-data commands that receive a spurious wake up

kdubb:
The fix here simply stops setting the command to complete after the `NoData` message from the server.  In `BindExecCommandImpl` this is never correct since we always send more messages after the `Describe` messages in this command.

Not setting the command status to complete causes the `waitFor` to keep waiting in the case of spurious thread wake ups. Previously, the command being marked complete, would cause `waitFor` to exit prematurely if it received a spurious wake up.


kdubb:
Fixed #252

PR #273, opened 2016-07-14
Housekeeper warning fixes

kdubb:
Hopefully these changes fix the last of the invalid warnings from the housekeeper.

Essentially the issue was that whenever the database connection was disconnected without explicitly calling `close`, via exception or other random networking/server event , the connection was left in an odd _not open_ but _not closed_ state.


brettwooldridge:
Awesome and encouraging activity on the project lately!

PR #292, opened 2016-12-14
Fix ByteBuf leaks using new Netty features

kdubb:
@jesperpedersen Can you confirm this works for you? I saw that 4.0.42 had issues.  This brings us to the latest netty series


jesperpedersen:
Netty 4.1.7 is needed to get the required fix

PR #293, opened 2017-01-05
escape ? operator with ??

pkolaric:
escape ? jsonb operator (used in postgresql 9.4+ https://www.postgresql.org/docs/9.4/static/functions-json.html) with double ??


jesperpedersen:
Could you add a test case for this too ?

pkolaric:
I've added the two additional operators that start with ? and a test case.

jesperpedersen:
Your test case needs
```java
    if (!((PGConnection)con).isServerMinimumVersion(9, 4))
      return;
```
in order for the Travis build to pass. Also, you could squash your commits ?

pkolaric:
Ah, completely missed that one... Thanx.

jesperpedersen:
Merged, thanks for your contribution !

Issue #195, opened 2015-09-02
Functions retrieval feature


adriens:
Hi, does your driver support JDBC functions retrieval (native jdbc driver does not for now).

Thank you in advance for your help.

Kind Regards,

Adrien

davecramer:
What do you mean by functions retrieval ?

Dave Cramer

adriens:
In fact, when i use schemacrawler on postgres... when it's performing queries on  connection, on functions i get the following log :

`ao¹t 30, 2015 5:49:19 PM schemacrawler.crawl.RoutineRetriever retrieveFunctions
AVERTISSEMENT: JDBC driver does not support retrieving functions
java.sql.SQLFeatureNotSupportedException: La fonction org.postgresql.jdbc4.Jdbc4
DatabaseMetaData.getFunction(String, String, String) n'est pas encore implÚmentÚ
e.
        at org.postgresql.Driver.notImplemented(Driver.java:670)
        at org.postgresql.jdbc4.AbstractJdbc4DatabaseMetaData.getFunctions(Abstr
actJdbc4DatabaseMetaData.java:87)
        at schemacrawler.crawl.RoutineRetriever.retrieveFunctions(RoutineRetriev
er.java:175)
        at schemacrawler.crawl.SchemaCrawler.crawlRoutines(SchemaCrawler.java:21
4)
        at schemacrawler.crawl.SchemaCrawler.crawl(SchemaCrawler.java:564)`

and in fact it's the driver that did not implement this feature.

kdubb:
@adriens That log is from the official Postgres JDBC driver, not ours.  That being said, neither driver supports "DatabaseMetaData.getFunctions". The reason is that Postgres doesn't have an actual implementation of SQL "functions".  It does have procedures and "getProcedures" is implemented in both drivers.

davecramer:
@adriens I just pushed a change into the official driver to support them,
but as Kevin points out it is just the same as getProcedures

Snapshots can be found on maven

Dave Cramer

jesperpedersen:
Providing an alias for getFunctions() -> getProcedures(), and getFunctionColumns() -> getProcedures() is a bad idea, since the ResultSet will contain different columns, and use different constants (which aren't binary compatible).

Throwing the SQLException is a valid "return" type for both functions, so tools needs to account for that.

As such, I'm closing this feature request, since PostgreSQL are using procedures, and those methods are implemented.

PR #298, opened 2017-01-16
[#195] Make getFunctions() and getFunctionColumns() aliases

jesperpedersen:
Will sleep on this one...

@kdubb Thoughts for inclusion ?


jesperpedersen:
Nah, bad idea

PR #304, opened 2017-02-07
Fixed Netty version range for pgjdbc OSGi bundle

oscerd:
Actually the project is based on Netty 4.1.x, the Netty Version range in maven-bundle-plugin configuration is wrong. It can't be installed in a OSGi container in this way. I guess this must be fixed, with a quick new release if possible.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -180,7 +180,7 @@
             <!-- do not import ourself -->
             <Import-Package>
               !com.impossibl.postgres*,
-              io.netty.*;version="[4.0,4.1)",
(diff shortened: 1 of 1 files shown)


kdubb:
Probably a good time to upgrade to Netty 4.1.8 then release a 0.7.1,

@jesperpedersen 4.1.8 is good to go correct?

oscerd:
From the OSGi pov there will be no problem with 4.1.8 :-)

oscerd:
By the way this PR seems to be ok

PR #287, opened 2016-10-10
Make BasicContext.loadLocale() handle Windows locales.

brcolow:
The locales that PostgreSQL uses on Windows are different from their
*nix counterparts. They look like 'English_United States.1252' whereas
a *nix locale looks like 'en_US.UTF-8'.

Attempting to parse a Windows locale with the current code results in
an error:

Caused by: java.util.IllformedLocaleException: Ill-formed region: United States [at index 0]
        at java.util.Locale$Builder.setRegion(Locale.java:2495)
        at com.impossibl.postgres.system.BasicContext.loadLocale(BasicContext.java:285)
        at com.impossibl.postgres.system.BasicContext.init(BasicContext.java:265)
        at com.impossibl.postgres.jdbc.PGConnectionImpl.init(PGConnectionImpl.java:249)
        at com.impossibl.postgres.jdbc.ConnectionUtil.createConnection(ConnectionUtil.java:182)
        at com.impossibl.postgres.jdbc.AbstractDataSource.createConnection(AbstractDataSource.java:747)
        at com.impossibl.postgres.jdbc.PGDataSource.getConnection(PGDataSource.java:66)
        ... 6 more

This is because Java is not able to parse a Windows locale. So, we
create a mapping from Windows to *nix locales in utils/Locales.java
and use the utility function getJavaCompatibleLocale(String winLocale)
to see if the locale returned by PostgreSQL matches any known Windows
locale.

The list of locales in utils/Locales.java is not exhaustive, but it
should be sufficient at least as a start.

PR #307, opened 2017-02-14
Fix NullPointerException on Windows

RoelVR:
On Windows, I got a NPE after upgrading to 0.7

The cause seems to be a small bug in BasicContext introduced in #287. This fixes the issue for me on Windows 10.

### DIFF
--- a/src/main/java/com/impossibl/postgres/system/BasicContext.java
+++ b/src/main/java/com/impossibl/postgres/system/BasicContext.java
@@ -287,8 +287,9 @@ public class BasicContext implements Context {
       }
 
       // Check if locale matches a win32 type locale (e.g. "English_United States.1252")
-      if (Locales.getJavaCompatibleLocale(localeSpec.split("\\.")[0]) != null) {
-        localeSpec = Locales.getJavaCompatibleLocale(localeSpec);
+      String windowsLocale = localeSpec.split("\\.")[0];
+      if (Locales.getJavaCompatibleLocale(windowsLocale) != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/impossibl/postgres/system/BasicContext.java
```diff
@@ -288,7 +288,7 @@ private void loadLocale() throws IOException, NoticeException {
 
       // Check if locale matches a win32 type locale (e.g. "English_United States.1252")
       if (Locales.getJavaCompatibleLocale(localeSpec.split("\\.")[0]) != null) {
-        localeSpec = Locales.getJavaCompatibleLocale(localeSpec);
+        localeSpec = Locales.getJavaCompatibleLocale(localeSpec.split("\\.")[0]);
```
gsmet: Hi,

Maybe extract that to a variable to avoid doing the operation twice (and making the same mistake again in the future)?

RoelVR:
Added a variable as suggested

jesperpedersen:
Merged, thanks for your contribution !

andi-livn:
Ha, after getting these NullPointerExceptions on a single developer's windows machine, we just hunted this bug down ourselves, and were about to suggest the same fix. Thanks gsmet, RoealVR and Jesper for your fix.

Is there a new build in sight that will include the fix? The latest version on maven central, 0.7.1, still has the bug.

Siedlerchr:
It would be nice, if you could provide a snapshot release of the library, as one of our [JabRef users](http://discourse.jabref.org/t/postgresql-connection-jabref-4-0-4-1-4-2/1113) experiences the same problem.

PR #308, opened 2017-03-02
Use correct package versions for OSGI bundle 

sesuncedu:
The OSGI package version numbers are incorrect with respect to past changes in backwards compatibility in maven versions 0.3 through 0.7.1 . 

These package versions have not always been bumped appropriately in past releases , 
Package version numbers have only been set since maven version 0.7, and were forced to match the maven version; for example, in maven version 0.7.1, minor (backwards compatible) changes were made to ```com.impossibl.postgres.jdbc```, but the minor version of the package was not increased.

This pull request add package-info.java files to all packages, containing OSGI Version annotations that have version numbers set to be safe with all prior OSGI bundles.  
Packages that have seen at least one major version number change since maven-0.3 have been set to package version 1.0.0.  
Packages that have not had any major api changes  since 0.3.0 have been set to version 0.7.1. 

The baseline goal for the felix bundle plugin has been set to execute at the verify stage.  The baseline used for comparison is the last release (maven-0.7.1).  
Warnings have not been suppressed;  thus there will be warnings that a number of package versions have been set too high. These warnings will stop after the next non-SNAPSHOT release.

The Bundle-Version and maven versions are independent of the package versions. It can be helpful to increase these versions to match the most significant change in any given package, but it's not as problematic.   

The imports section of the package has been altered to self-import, and to pick up the version range for netty from the netty bundles (which happen to match the versions being explicitly set).  Other imports have been left unaltered, including some  which may cause minor issues under jdk 9 (e.g. ```javax.transaction```, which is supplied with jdk9, but which has to be added to the jigsaw module path explicitly.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,7 +158,6 @@
         <artifactId>maven-surefire-plugin</artifactId>
         <version>2.18.1</version>
         <configuration>
-          <skip>true</skip>
           <argLine>-Dio.netty.leakDetection.level=advanced</argLine>
           <excludes>
             <exclude>**/PerformanceTest.java</exclude>
--- a/pom.xml
+++ b/pom.xml
@@ -89,6 +89,12 @@
       <version>${netty.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.osgi</groupId>
+      <artifactId>org.osgi.annotation</artifactId>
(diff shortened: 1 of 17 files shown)


sesuncedu:
Closed, replaced with #310

Issue #309, opened 2017-03-02
OSGI package  version numbers over / under optimistic 


sesuncedu:
The OSGI package version numbers in releases of pgjdbc to date have been  over and under optimistic, which I think means the glass is full, but it's only a half. 

Some packages have not had visible API changes since the first release with bundle headers (0.3), but most have had at least one non-backwards compatible change. 

Versions 0.7 was the first release with non 0 version numbers - all 0.7.0.
Version 0.7.1 bumped every package version  to 0.7.1. In most cases this was unnecessary, but for .jdbc, there were minor level changes that should have given that package a version of 0.8.0 or higher. 

I have submitted a PR that adds a package-info.java to each package with a safe version number. 
Packages with unchanged APIs since jar version 0.3 have been given version 0.7.1. Other packages, with at least one major API change since 0.3 have been given version 1.0.0. 
This is safe, since there have been no packages versions with a major > 0.

I have also set up the Felix bundle plugins baseline goal, attached to the  verify phase. It will issue warnings about version numbers being increased too much, because it is using the latest release as the baseline version, so doesn't know about past problems.  These warnings will stop if a release is made with the new package versions, and are not set to break the build. 
Changes that affect the api will cause errors, with suggestions for replacement package versions. The Felix output displays any api diff in a friendlier fashion than the bnd plugin. 

Whilst I was changing the bundle plugin settings, I also removed the explicit versioning for netty-io, as the import versions are correctly  generated from the dependencies. I also enabled self-import of pgjdbc packages, which can be useful at times. 

I did not make any changes to which packages are exported ; it can be useful to avoid exposing private packages. 
I also did not change the settings of any JRE imports to make them optional. Thus the bundle cannot be resolved  if the javax.xa packages aren't available (this will be an issue with jigsaw, which hides them by default). 
But then again, everything is an issue with jigsaw.

kdubb:
Replaced with #372

PR #310, opened 2017-03-02
Fix OSGI versions, Per #309

sesuncedu:
As described in issue #309.
Differs from PR #308 in that commit to fix for accidental test skip  is squashed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -93,6 +93,8 @@
       <groupId>org.osgi</groupId>
       <artifactId>org.osgi.annotation</artifactId>
       <version>6.0.0</version>
+      <!--compile time only -->
+      <scope>provided</scope>
     </dependency>
 
   </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -89,6 +89,12 @@
       <version>${netty.version}</version>
     </dependency>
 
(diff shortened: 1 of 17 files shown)


jesperpedersen:
I don't think we want to have the OSGi dependency; it must be possible to fix this issue some other way.

Only the JDBC spec needs to be well-defined between releases; all other interfaces are specific, and may change between versions. IMHO, if that means verion bumps even if those interfaces doesn't change => so be it

kdubb:
@jesperpedersen Do we know what the preferred/best method to manage OSGI packing is? I'm ok with adding dependency as runtime/provided because missing annotation classes do not preclude a class from being loaded.

Is moving to Java9 and generating modules an easier path? I assume the Java9 module system manages/monitors the export of packages from modules (or tools exist that can) and OSGI can be piggybacked on that.

jesperpedersen:
@kdubb I think keeping base as JDK8 is more important than OSGi descriptors. Frankly, I'm not seeing OSGi deployments anymore... So, runtime / provided is ok with me.

Once JDBC updates then we can follow what ever they did with modules and so on.

kdubb:
Replaced with #372

PR #313, opened 2017-03-03
Handle null values passed to setters for primitive values 

sesuncedu:
Method property setter generates (and masks) illegal argument exceptions when called with a null value for a primitive property. This leave the value unchanged,but generates an exception with a full stack trace. 

This Change records whether the argument type of the method is a primitive, and does not invoke the method if the value is null.


kdubb:
Fixed in #349

PR #322, opened 2017-03-16
Fix IllformedLocaleException connecting to DB with a Norwegian locale

pidue:
Since version 0.7.1 we were unable to connect to a PostgreSQL cluster if created with Norwegian Nynorsk or Bokmål locales. 

`java.util.IllformedLocaleException: Invalid subtag: Norwegian Bokm?l [at index 0]
  at java.util.Locale$Builder.setLanguageTag(Unknown Source)
  at com.impossibl.postgres.system.BasicContext.loadLocale(BasicContext:299)
  at com.impossibl.postgres.system.BasicContext.init(BasicContext:273)
...
`

In the LOCALES lookup table (com.impossibl.postgres.utils.Locales) the two Norwegian locales (Nynorsk and Bokmål) are missing. 

The fix required to support Nynorsk locale is straightforward, but things get complicated with Bokmål. Unluckily, it's the only known locale with a non-ascii name, so the query in BasicContext.loadLocale returns a wrongly encoded locale name and a table lookup is not possible. A prefix match is the only solution we found.

### DIFF
--- a/.travis/travis-install-dependencies
+++ b/.travis/travis-install-dependencies
@@ -34,7 +34,7 @@ sudo tail /var/log/postgresql/postgresql-${PGVERSION}-main.log
 
 # Setup postgres' users and databases
 sudo -u postgres psql -c "CREATE USER pgjdbc WITH PASSWORD 'test';"
-sudo -u postgres psql -c 'CREATE DATABASE test WITH OWNER pgjdbc;'
+sudo -u postgres psql -c 'CREATE DATABASE test WITH OWNER pgjdbc LC_COLLATE 'Norwegian Bokmål' LC_CTYPE 'Norwegian Bokmål' ;'
 
 # Install hstore extension if >= 9.1
 if [ "${PGVERSION}" != '8.4' -a "${PGVERSION}" != '9.0' ]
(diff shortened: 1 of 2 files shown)


pidue:
We were unable to submit a proper test case, due to the lack of understanding of your test infrastructure. We can provide the createdb options to setup a new db that causes the IllformedLocaleException, but we didn't find how to integrate this in the Travis install script.

`initdb —locale="Norwegian Bokmål" -E UTF8 -U postgres data`

pidue:
Only the Windows version of postgresql causes this issue, because of the different locale descriptors format. 

On Linux and OSX the two locales are called respectively nn_NO and nb_NO,  so they are compatible with Java locale naming standard.

There is no way to create a db with the locale descriptors involved in the bug on a *nix OS, nor testing in a Windows environment on Travis, so I can't see how to test this PR...

kdubb:
This locale is problematic and has been deprecated in PostgreSQL (details [here](
https://wiki.postgresql.org/wiki/Changes_To_Norwegian_Locale))

On top of this a new method of matching locales that doesn't require lookup tables is now being used. Implemented in 4acedf496ccb439a143f861e3124dd8673574dfd

This should be well taken care of going forward.

Tostino:
A.    <./.    Aaa    c.  C.      .

PR #334, opened 2017-09-12
Changes the URL regular expression to allow the use of any character

mlillywhite:
In postgresql, there are no restrictions on the database name if it's enclosed in double quotes; we use dotted notation for some database names, which was not supported by the driver.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,7 +46,7 @@
   </parent>
 
   <properties>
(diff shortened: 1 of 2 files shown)


kdubb:
This will be included in #357 since it had to be changed for sockets. It includes a "small" difference in that it doesn't allow slashes as that causes a bit of havoc; so it will be "anything but ?, & or /".

PR #335, opened 2017-09-18
Fixed NullPointerException on getConnection() against server v9.6.5.

javamachr:
If server was underload load attributes query contained temporary tables which caused Unknown object OIDxxxxxxxxxxx error from server resulting in NullPointereException being thrown on getConnection() call. This commit adds condition to omit temporary tables in the call.


kdubb:
Verified it works with all current versions... added to general query rather than a specific query in PR #349

PR #349, opened 2018-12-10
Rework

kdubb:
After, literally years, I've finished the conversion of the driver to buffer based data management.

This has many advantages including easier to implement encoding and decoding, better memory management, and better conformance to mainstream driver in certain areas (e.g. dates/times/timestamps).

Essentially, instead of immediately decoding each row's column into Java objects and then "coercing" those objects to whatever type was requested via the JDBC API (e.g. getString), the row data is now stored as a Netty buffer and decoded when the API request is made.
Conversely, parameter data is stored in Netty buffers which means the encoder is called during the API call (e.g. setString) and the value is encoded immediately.

Altogether this means encoding and decoding has more meaningful information to work with, which ensures the conversions are "correct"; and if they happen to not be correct we have all the required information to fix the problems.

Some byproducts of these changes are better performance and much less garbage created during normal API use.


jesperpedersen:
Great, I'll take a look at this during the week

pgjdbc-ng.iml
```diff
@@ -9,13 +9,13 @@
           <property key="Bundle-DocURL" value="http://www.impossibl.com" />
           <property key="Import-Package" value="!com.impossibl.postgres*,io.netty.*;version=&quot;[4.0,4.1)&quot;,*" />
           <property key="Bundle-Name" value="PGJDBC-NG" />
-          <property key="Include-Resource" value="META-INF/services/com.impossibl.postgres.system.procs.ProcProvider=$MODULE_DIR$/src/main/resources/META-INF/services/com.impossibl.postgres.system.procs.ProcProvider,META-INF/services/java.sql.Driver=$MODULE_DIR$/src/main/resources/META-INF/services/java.sql.Driver" />
+          <property key="Include-Resource" value="META-INF/services/com.impossibl.postgres.system.procs.ProcProvider=/Users/kdubb/Projects/pgjdbc-ng/src/main/resources/META-INF/services/com.impossibl.postgres.system.procs.ProcProvider,META-INF/services/java.sql.Driver=/Users/kdubb/Projects/pgjdbc-ng/src/main/resources/META-INF/services/java.sql.Driver" />
```
davecramer: @kdubb this has your personal home dir in it.

kdubb: Thanks... I really do hate these files...

jesperpedersen:
Using https://github.com/jesperpedersen/postgres-tools/tree/master/SQLLoadGenerator and doing a run with https://github.com/jesperpedersen/postgres-tools/tree/master/Replay on a 500Mb database and executing 1000000 SQL statements gives

```
42.2.5: 10130 ms
0.7: 31825 ms
Rework: 83571 ms
```

Replay is executed with the `-r` switch. PostgreSQL 11.x w/ default settings for all runs.

More investigation needed.

jesperpedersen:
With `housekeeper=false` - 0.7: `27443 ms` and rework: `77559 ms`

kdubb:
@jesperpedersen The statement cache wasn't implemented correctly, a `hashCode` was improperly implemented. The result was that after filling the cache to its max size it then un-prepared and re-prepared every statement that was executed; making it much worse than just a disabled cache.

I've just got SQLLoadGenerator and Replay working and am going to continue to investigate any more major performance issues and report back.

jesperpedersen:
Thanks ! That change gives `27151 ms` for rework.

kdubb:
I've discovered that the explicit `BEGIN` appears to be one of the major factors. Commenting out the transaction functionality in Replay, thus using auto commit, drops my timing in from ~20000 ms to ~13000 ms.

I'm switching the code to prepend `BEGIN` to relevant queries instead of waiting for a complete roundtrip of an isolated `BEGIN` query.

jesperpedersen:
What about interactions which are empty ? E.g. only contains `BEGIN` and `COMMIT` because the client never issues any JDBC commands; a container may issue the `setAutoCommit(false)` and the `c.commit()` calls.

kdubb:
Calling `commit` on a connection is a query and it would detect and, somewhat ironically, prepend a 'BEGIN' query.

kdubb:
Another option could be to detect the `IDLE` transaction state in `Connection.commit` and do nothing.

jesperpedersen:
I would say latter is likely the best, due to performance.

In case of issuing `BEGIN; XYZ` we will have to make sure that error scenarios for both statements are handled correctly - see `PGXAConnection::start(Xid, flags)`

kdubb:
@jesperpedersen According to my latest runs, this PR is within 15% of the mainline driver.  After hiding transaction roundtrips I discovered numerous superfluous "close" commands roundtrips sent to the server unnecessarily.

FYI, superfluous `COMMIT` commands have been ignored for a long time (setting to manual and then setting to auto commit). So I see no issues with burying them as they are now.

Finally, I have another branch I'll PR after this closes that shares the registry. between connections vended from a datasource.  This brings our performance in that scenario even closer because only connection does the type loading. We'll discuss the merits and possible issues in that PR.

I think this PR is close. It's extremly better than our 0.7.1 at this point in terms of performance and is way more conformant with mainline.

davecramer:
@kdubb so I'm a little disappointed that this isn't *better* than the mainline driver. My hope was that we'd see significant improvement.

kdubb:
@davecramer I'm still really just attacking the low hanging fruit. I would say though that a significant performance improvement would be relative. As you are keenly aware, a significant amount of of work has been done on the mainstream driver to bring it to its current state. Anything new was never going to torch it in a race and certainly not without a real investment to make it happen. I'm hopeful we'll get there.

jesperpedersen:
Latest changes gives `20285ms`.

I'll look at the code changes.

jesperpedersen (changes requested):
Overall it looks good. Maybe squash the changes into one commit such that all changes are co-located.

* `PGStruct::finalize()` isn't ideal, maybe we could tie it to `Connection::close` too in order to cleanup when one of those happen

* The use of lambda's requires extra meta space, and have therefore been banned in some projects, such as WildFly, due to usage in memory constrained environments so I think we should remove usage of those in the driver

kdubb:
@jesperpedersen  I removed the finalizer by switching to allocating structs in unpooled buffers. One of the my "to do" items is going to be investing the removal of pooled byte buffers altogether.  Netty5 is looking at use Java 9 "Cleaner" (it's basically identical to our `Housekeeper`, including using phantom references and all). In my following comments about performance it might just not be necessary to worry about it; maybe even use the Housekeeper to cleanup all pooled buffers or switch to Java 9 and use a `Cleaner`.

As far as lambda's go I'm going to merge this PR with them in.  I'll create a "to do" issue to discuss the merits of their removal (I have questions... like why are anonymous classes allowed). We can take that up in that issue.

Soon I'm going to squash and merge 😀

kdubb:
@jesperpedersen @davecramer 

After days of YourKit, WireShark, investigation, complete rewrites, hair pulling, etc... I'm back with news about performance...

** The Bad **
We're not going to beat a synchronous blocking socket with Java's NIO. It just 'ain't gonna happen'... I finally faithfully replicated @jesperpedersen's results using `Replay` and `ng` is _now_ around 17-18 seconds vs mainline being around 11-12 for my testing setup.

I'm satisfied I've investigated every possible avenue, including replacing Netty with my own async IO, replacing it with a single threaded async IO, replacing it with single threaded synchronous IO and finally just using a stinking socket and reading/writing directly to it with no threads involved.  Low and behold the socket is far and away the fastest option. This is supported by just about every source on the internet.  Even with a thrown together 10 minute hack of an implementation `ng` is reduced to 13-14 seconds.

Some of our problems were due to issues with the way messages, and what messages, were sent to the server. WireShark helped me bring a level of parity between them and showed some improvement areas. But even if we optimize every last bit we're not going to make up that ground...

** The Good (well great) **
We're probably "testing it wrong"...  The testing rig I was using (and apparently @jesperpedersen too) was testing a localhost connection to a localhost database. As a matter of fact I was originally using docker to host postgres but that gave me completely different results from @jesperpedersen. When I switched to a native server running on my machine I began to see the same results.

Then, after all the tuning I could do I thought I test it in a more "real world" scenario.  I have to fairly new iMacs on my desk one not used at all. I connected them via FireWire 3 and ran Postgres on the idle machine.  The performance disparity almost completely disappeared. Short version... when running in a pretty optimal scenario the async's overhead is negligible.

These tests were done using a rewrite of the underlying protocol that is completely asynchronous, which was a goal of mine during this because I want to implement ADBA on top of the engine we have. I am opening a new PR immediately after I merge this one so we can discuss that there.

For me all option are on the table. Keeping Netty, removing it, replacing it with another framework or something homegrown. I think the real key is that asynchronous is a requirement moving forward and we need to figure out the best possible solution.

PR #350, opened 2018-12-19
Rework 2 -  Async Boogaloo

kdubb:
This is a follow on to #349 that replaces the entire protocol layer with a "proper" asynchronous request submission channel.

The server connection is now an asynchronous FIFO queue. Multiple requests can be submitted __without__ waiting for completion and the server responds in turn.  The connection maintains a stack of protocol handlers that determine when and how each group of commands is considered "complete".  If any command receives an error from the server, the framework fails all following requests up until it reaches a `sync` message... as the protocol was designed.

JDBC is synchronous but upcoming specs are tackling that, specifically ADBA.  This will allow us to support both API's on on core.

The issue of performance has now been investigated deeply on top of #349.  There are serious concerns about Netty vs blocking I/O.  It is my hope to get this PR completed and merged and then investigate the most performant I/O method that can support an asynchronous driver.

Other changes in this PR:
* Statement description cache
  * Caches server descriptions of queries to ensure the correct type information is being used. It can borrow a description from the prepared statement cache if one is available.
* Prepared statement heat caching
  * Previously all statements were "pre" prepared when the cache was enabled. Now a threshold can be set to enable 
* WireShark style protocol tracing
  * Turning it on dumps a succinct message trace inspired by WireShark; with message routing information intermixed. It makes debugging and performance investigation _easy_.


kdubb:
I forgot to mention one JDBC feature can really benefit from this, PreparedStatement batches. When re-implemented using this request execution model we can send all the requests and then process each response as it comes in.  Should be pretty cool.

davecramer:
@kdubb yes, that is very cool!

jesperpedersen:
I was reading this when you did the merge, but it looked good :)

kdubb:
Apologies... I may have been a bit too eager to merge these as I want to get into develop so that snapshots are pushed. I'll wait for your review on the others; big changes, definitely want input.

PR #354, opened 2018-12-23
Rework 3 - Fast Registry

kdubb:
This PR builds on #350 to make the registry less of a burden.  It's now much faster at initializing and shares the information between connections when possible.

This commit reduced the test suite time on my local machine from an average of ~72 seconds down to ~28 seconds.

This doesn't change the way the driver works besides delaying the loading of type information until needed and speeding things up.  This keeps it compatible with certain extension types that have non stable type oids.

It's fast enough at this point I'm not sure anymore changes are really necessary but, if warranted, future commits could lessen the requirement on the registry altogether while still supporting binary & text transfer of `SQLData`, `Struct`, `Array`, etc. 

Three major things were changed:
* Initialization only fetches a small number of types
  * Went from thousands of types to around 170 for a default test database (has some extensions like `hstore` installed.
* Drastically reduced the amount of data downloaded by the driver
  * Optimized queries to fetch only required data.
* Implements registry sharing
  * Registries can only be shared between connections created by the same `PGDriver` or `PGDatasource` and they must have the exact same set of properties; it _should_ ensure complete compatibility.  All performance based applications will be using connection pooling from a datasource so this should be a big win.


jesperpedersen (approved):
I think this looks good - feel free to squash and merge.

PR #355, opened 2018-12-30
Rework 4 - As in JDBC 4 (.2)

kdubb:
Implements the JDBC 4.2 methods. Most implementations are trivial; implementing or routing to a large method using long instead of int.

The major changes are with regard to type specification...

`PGType` is an enum that allows specifying exact PostgreSQL types via the 4.2 `SQLType` API mechanism. When calling `setObject` or `updateObject` an exact, non-standard, type can be specified (e.g. `text`, `uuid` or `money`).

`PGConnection.resolveType(String)` allows the user to resolve any named type to a usable `SQLType` that will not require the driver to do a continuous name lookup; this is required to ensure that the client can synchronize with any DDL commands that might change type information mappings.


jesperpedersen (changes requested):
Batch updates are a bit weird, so BatchExecuteTest.java is correct. Therefore changes are needed to keep that logic.

https://github.com/eclipse-ee4j/jakartaee-tck hasn't been updated to the new JDBC specs, but the basic Batch test cases are in there.

Also, PostgreSQL 11 could be added to the Travis matrix.

kdubb:
@jesperpedersen Are you saying the our BatchExecuteTest.java is correct or incorrect in this PR?

I'm going to add 11 to the travis matrix once I figure out the easiest way to do it; which probably requires refactoring the configuration. The changes regarding travis in this PR were only to fix the fact it was using 9.6 instead of the listed version.

jesperpedersen:
Incorrect. The test case needs to stay the way it is.

kdubb:
From javadoc for `Statement.executeBatch()`...
> A value of <code>EXECUTE_FAILED</code> -- indicates that the command failed
> to execute successfully and occurs only if a driver continues to
> process commands after a command fails
>
> The possible implementations and return values have been modified in
> the Java 2 SDK, Standard Edition, version 1.3 to
> accommodate the option of continuing to process commands in a batch
> update after a <code>BatchUpdateException</code> object has been thrown.

PostgreSQL bails at the first error it encounters... It seems that we should never be adding `EXECUTE_FAILED` because of that.  According to this it seems that any driver throwing a `BatchExcecutionException` exception should not include an `EXECUTE_FAILED` code.

jesperpedersen:
Unfortunately, the TCK doesn't work that way, and no development on the JDBC part is done atm. I agree that this is weird, hence the added `strictMode` parameter and the explicit test cases.

jesperpedersen:
Thanks. Feel free to squash and merge.

Issue #339, opened 2018-02-13
What if the connection is lost ?


d3bt3ch:
I am registering a listener with the database server for notifications. What if there is a network failure or the database server is rebooted ?

vaIgarashi:
bump

stodge:
I am also interested in knowing this.

davecramer:
The connection drops and you do not get notified.

kdubb:
As of #356, `PGNotificationListener` has a `closed` method to notify listeners when the connection is closed; you can now make a new connection or handle it some other way.

Issue #318, opened 2017-03-06
PGNotificationListener extends NotificationListener instead of vice versa


sesuncedu:
The API class ```com.impossibl.postgres.api.jdbc.PGNotificationListener``` is defined as: 
```java
import com.impossibl.postgres.system.NotificationListener;
[...]
public interface PGNotificationListener extends NotificationListener {
}
```
Where ```com.impossibl.postgres.system.NotificationListener```  is defined as 

```java
public interface NotificationListener {
  void notification(int processId, String channelName, String payload);
}
```

This requires  ```NotificationListener``` to be exported as part of the public API. 
The declaration of ```NotificationListener::notification``` could be moved to PGNotificationListener, and NotificationListener could be defined as extending PGNotificationListener. Alternatively, all references to NotificationListener could be changed to refer to NotificationListener.

kdubb:
`NotificationListener` has been removed in #356

PR #356, opened 2019-01-03
Enhance asynchronous notifications

kdubb:
Removes useless base interface and adds a `closed` notification to allow listeners to know when no more notifications are coming.

Solves #339 & implements #318


jesperpedersen:
LGTM

PR #357, opened 2019-01-04
Unix socket support

kdubb:
Adds support for unix domain sockets via Netty.

* Added `unixsocket` URL property to add a domain socket to the list of hosts.
  * Unix socket addresses are _prepended_ to the attempted address list; an assumption was made that if you are using a unix socket with TCP fallback the unix socket is preferred.
* Requires Netty native libraries to be on the classpath
  * Can be done by customizing `pom.xml` (see netty docs) but is not done by default.
* Needs to be added to travis configuration


kdubb:
**NOTE**

The `unixsocket` property can be a directory or a specific unix domain socket file.  If the property points to a directory it attempts to locate a matching `.s.PGSQL.XXXX` file automatically.  If it finds multiple files it chooses the first one it finds (and logs a warning).

jesperpedersen:
Looks good, apart from your path in `pgjdbc-ng.iml` and missing update to `README.md`

PR #358, opened 2019-01-04
Pull driver name from JAR manifest (like version)

kdubb:
This sets the driver name via `Package.getImplementationTitle` the same way the version is now applied.

This has the side effect of changing the `project.name` to match what the driver name has always been.  AFAIK changing `project.name` has no ramifications on uploads to maven central.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -262,7 +262,7 @@
     </profile>
   </profiles>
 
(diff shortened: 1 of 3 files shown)


jesperpedersen (approved):
LGTM

PR #359, opened 2019-01-04
Travis configuration updates

kdubb:


jesperpedersen:
Why the change of the test user name ?

kdubb:
A bout of laziness... the SSL tests currently require the name `test` because that's how the certificates were generated; instead of regenerating them I switched the name. I planned on switching back but with the new `./start-test-container.sh` spinning up a perfect docker test target of any version it seems meaningless; if it's warranted I will do it.

kdubb:
I'm thinking of replacing the `./start-test-container.sh` script with a maven docker plugin that'll boot the container in the same way for integration tests.  That would make testing as simple as having docker installed and issuing `mvn verify`. Any reason not to do that? 

I can only see positives... I don't have to start a local test server, it always testing a pristine database, testing against different versions will be as easy as setting a variable, etc.  Am I missing something negative?

jesperpedersen:
I'm ok with the Travis changes, but for the build itself I don't think we should rely on docker for the test cases. Having a profile that uses docker is ok.

I'm using https://podman.io/ instead of docker, f.ex.

kdubb:
@jesperpedersen  I decided to leave maven alone. The start script is usable locally and for travis. Once checks complete I'm merging.

kdubb:
This merge brought the first full test of `develop` against all actively supported versions of PostgreSQL and it deployed the first `0.8-SNAPSHOT` release with all the recent changes 🎉🎉🎉

Issue #315, opened 2017-03-06
DataSource configuration issues:


sesuncedu:
1.  The name of the standard datasource property for storing "name of a particular database on a server" is "databaseName"; pgjdbc-ng uses "database". 
2. The name of the standard datasource property for storing the "database server name" is "serverName"; pgjdbc-ng uses "host".
3. The name of the standard datasource property for storing the "port number where a server is listening for requests" is "portNumber"; pgjdbc-ng uses "port".
4. pgjdbc-ng does not support the "description" property, which is apparently the only required property. 

  (JDBC 4.2 specification, §9.6.2 DataSource Properties) 

Additionally, most  drivers (including pgjdbc-TOS)  support the "url" property.

kdubb:
Fixed in #360

PR #360, opened 2019-01-06
Revamp how settings are defined, found & stored

kdubb:
Added a `Setting` object which defines information related to each setting (description, name, alternate names, default value, etc).  Settings can even define a dynamic source for it’s default value. This will ensure that we always know everything about each setting we define and where/how the default values are determined.  Removes a lot of parameter/setting copying code to make sure things are named right and in the right properties instance.

Also, added `Settings` which is much like `java.util.Properties` in that it stores everything in text but it allows access via `Setting` definitions.  `Settings` logs warnings for unrecognized names and logs errors for invalid values that are stored.

See `JDBCSettings` (previously `PGSettings`) & `SystemSettings` to see how the definitions have changed.

Allowing alternate names for a setting allows changing them in the future. This isn’t just a “nice to have” currently we use incorrect names for datasource properties; this allows a backwards compatible fix for issues like #315.


kdubb:
@jesperpedersen In the recent commit I made a decision to require that all settings have a primary name in lowercase dot-dash format (e.g. `this.is.a.property-name`).  I generally prefer this format for settings because they seem easier to remember and type correctly without intermixed case.

The dotted format is easily convertible to a JavaBean property name (e.g.  it becomes `thisIsAPropertyName`), which I rely upon in the new `AbstractDataSource` implementation. It's also used in the new associated test that ensures all settings from the related setting groups are available on `AbstractDataSource` as bean properties.

I generally like this change, but it means going forward all properties must have a primary name matching this format.  Wanted to get outside input before deciding.

jesperpedersen (dismissed):
I'm ok with these changes.

However, `README.md` needs to be updated to document the default properties

kdubb:
I'm working on auto-generating a properties document from the settings and going to link to it from `README.md`

jesperpedersen:
Ok, cool

PR #361, opened 2019-01-09
Cleanup JDBC/Type mapping

kdubb:
* Remove `PrimitiveType` completely, types are now directly mapped to JDBC type using `PGType`
* Replaces all use of `java.sql.Types` in the mapping process with `SQLType`
* Implements some missed JDBC 4.2 methods in `PGCallableStatement` that use `SQLType` mapping
* Uses `PGType` to produce data for `JDBCTypeMetaData`
* Moves all `Object` to `SQLType` discovery to `JDBCTypeMapping`


kdubb:
It's a large charge only in that it touches a lot of files. Each Decoder was required to implement a method that provided its `PrimitiveType` but this enum was inaccurate and not needed after rework; so it was removed.

jesperpedersen:
There is something wrong with skipping the SSL tests, if the server is started manually with no SSL support

jesperpedersen:
Using the full setup (hstore, citext, sslinfo) with all the databases, and the tests passes.

LGTM

jesperpedersen:
Something to note; PostgreSQL 12 will remove support for `WITH OIDS`

kdubb:
`WITH OIDS` only affects non system tables correct? I'm assuming the system tables will all retain their `oid` column, or will it be renamed?

jesperpedersen:
Just user tables; and a quick look at the 2 test cases doesn't really show anything that can't be removed

PR #362, opened 2019-01-10
Generated DataSource & Docs, Multi module maven

kdubb:
* Generate `AbstractGeneratedDataSource` from global settings groups.
  * Includes JavaDoc for each property generated from settings.
* Generate `SETTINGS.md` documentation which details each available parameter and the way(s) it can be provided.
* Switch to multi-module maven project
  * Code and Doc generation is from annotation processing which needs to be built before the driver is compiled. It was only acheivable via modules.
  * I tried numerous tricks to not have to force moving the files to a subdirectory but none worked.


jesperpedersen:
LGTM.

I'm getting a diff on `SETTINGS.md` after `mvn clean package`, so that file needs to be updated before the merge.

kdubb:
@jesperpedersen Even with the latest commit adding ranges? It was out of date before that but I don't get a diff after `mvn clean package` with f8a88333b9f86c040782d2553419780b2144435e

jesperpedersen:
Yes, that was the version I looked at / tested

kdubb:
I can't seem to get a diff.  This reassures my fears with generating files into the repo.  We should generate and deploy to `gh_pages` or something.

jesperpedersen:
Could be a Maven plugin thing. However, I say just commit what you have such that `SETTINGS.md` is visible on `develop`.

PR #363, opened 2019-01-12
JDBC API level tracing

kdubb:
This completes the picture for tracing with the driver; JDBC, SQL & Protocol tracing are all supported and each can be enabled/disabled and routed to files independently.

To support API tracing I added `spy`, a (mostly) auto-generated (via `spy-gen`) JDBC API spying library that wraps & unwraps API objects as they exit/enter the API. The library can actually be used with any driver and is capable of doing quite a bit more than tracing/logging.  It’s pretty complete as the entire test suite passes with API tracing enabled; only changes required were to use `unwrap` instead of casting connections (see previous commit).

Since `spy` is generated from the JDBC API interfaces and the JDK is helpfully compiled _without_ parameter names I added `jdbc-api` module that contains the relevant `java.sql` and `javax.sql` interfaces from JDK8 source and compiles them with parameter names enabled; `spy-gen` is run with these classes on the bootclasspath to generate traces with valid parameter names. As noted in the library's README, it is not in any way included in the package code.


jesperpedersen:
LGTM

PR #364, opened 2019-01-12
Make `strict-mode` strict again

kdubb:
Until now enabling `strict-mode` made the driver _less_ compliant. It’s fixed and `strict-mode` is now on by default.  Also, I’ve documented in the setting what behavior changes and when.

#### getColumnName/getColumnLabel
The only place strict mode was still used was in the discrepency between `ResultSetMetaData`’s `getColumnName` and `getColumnLabel`.  According to the spec `getColumnName` should return the table name (other drivers seem to agree at this point).  We had it reversed, strict mode made it return a column label first (if available) then fell back to the table name. Running in `strict-mode` was  “more compatible” / “less strict”.

I’ve now corrected this case and added full tests for both cases. I’v also enabled `strict-mode` by default. This shouldn’t be an issue because it was disabled before the logic was inverted.

#### BatchUpdateException
Along with fixing the previous case I added back another area I knowing defied the standard with regard to `BatchUpdateException` including and EXECUTION_FAILED flag.  In this case we were just “non-compliant” and included the flag. Since the spec is clear about this I re-instated the compliant behavior and stopped adding the flags when in `strict-mode` and add the flag when `strict-mode` is disabled. This change _does_ represent a reversal as, per default, it is now compliant.

### DIFF
--- a/SETTINGS.md
+++ b/SETTINGS.md
@@ -107,13 +107,20 @@ Default: `0`
 
 Enable or disable strict adherence to JDBC specification.
 
+Affected behavior when disabled:
+<ul>
+  <li><code>ResultSetMetaData.getColumnName(int)</code> returns a value equivalent to <code>ResultSetMetaData.getColumnLabel(int)</code>, if a label is available.</li>
+  <li>The <code>Statement.executeBatch(...)</code> family of methods insert an extraneous EXECUTE_FAILED status into <code>BatchUpdateException.getUpdateCounts()</code> even though PostgreSQL stops executing at the first error.</li>
+</ul>
+
+
 Driver Property: `strict-mode`
 
 DataSource: `getStrictMode()`/`setStrictMode(java.lang.Boolean)`
 
 System Property: `pgjdbc.strict-mode`
 
-Default: `false`
+Default: `true`
 
 #### Fetch Size
 
--- a/driver/src/main/java/com/impossibl/postgres/jdbc/BatchResults.java
+++ b/driver/src/main/java/com/impossibl/postgres/jdbc/BatchResults.java
@@ -55,6 +55,11 @@ interface BatchResults {
 class IntegerBatchResults implements BatchResults {
 
   int[] counts = new int[0];
+  boolean strict;
+
+  IntegerBatchResults(boolean strict) {
+    this.strict = strict;
+  }
 
   @Override
   public void setBatchSize(int size) {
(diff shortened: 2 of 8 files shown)


kdubb:
FYI, I did a little googling about the `getColumnName`/`getColumnLabel` debate.  It seems others agree the column name should refer to the actual table name if available.

https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-installing-upgrading-5-1.html

kdubb:
FYI, I did a little googling about the `getColumnName`/`getColumnLabel` debate. It seems others agree that the column name should refer to the table name.

https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-installing-upgrading-5-1.html

jesperpedersen (commented):
`StrictMode` [1] was added due to the JDBC TCK [2] [3], and was needed in order to pass with a setting of `true`.

So, you flipped the meaning. We want the "old" `StrictMode` setting of `false` to be the default, as it aligns better to the actual documentation of the JDBC API.

[1] https://github.com/impossibl/pgjdbc-ng/wiki/StrictMode
[2] https://github.com/eclipse-ee4j/jakartaee-tck
[3] https://github.com/eclipse-ee4j/jakartaee-tck/tree/master/src/com/sun/ts/tests/jdbc

kdubb:
@jesperpedersen From what I can tell it doesn't align better to the actual _documentation_ as much as it aligns better to the TCK. Clearly they should be in sync but if the TCK is expecting `getColumnName` & `getColumnLabel` are to return the same value then that seems to contradict the documentation.  Same with `executeBatch`.

It seemed to to me that `strict-mode` should align with the documentation of the API which is why I flipped it. Unless I'm misinterpreting the documentation; which may be the case.

jesperpedersen:
I called it `StrictMode` because TCK is the official "test suite".

I'm ok with flipping the definition as long as the wiki is updated, but maybe remove `StrictMode` and add a new property (`Jarkarta`) with default of `false`.

Ideally, we shouldn't need this mode, but TCK != documentation -- maybe Lance Anderson will update the TCK.

In short, I'm ok with what you decides on properties and their values.

kdubb:
I actually kind of like `strict-mode` at this point because it now aligns to the documentation; which is how people are learning to use it.

jesperpedersen:
Sure

PR #365, opened 2019-01-15
MetaData fixes

kdubb:
The spec defines meta-data result-sets column names to be returned uppercase. This ensures that all our retured columns meet spec. Column label comparisons were added to metadata resultsets to validate the names.

Also, fixes a couple typo in column names and adds some missing columns


jesperpedersen:
LGTM

PR #366, opened 2019-01-15
Type Fixes

kdubb:
A related set of fixes to the types, adds missing `maccaddr8`, removes deprecated types from `PGType` and adds the default java class it's mapped to.

Finally it enforces that all types implementing both text & binary return the same `defaultClass` property to ensure stability parity between text & binary.


jesperpedersen:
LGTM

PR #369, opened 2019-01-25
Switch to Gradle build

kdubb:
The switch was made because maven’s support for building a generator library and/or annotation processor in the same project is lacking and causes update problems. Gradle has `buildSrc` and is designed to work with this exact scenario.

### Docker
Docker support has been added to the build scripts. It will automatically test against a “default” version of PostgreSQL (currently 11) via the normal Gradle `test` or `build` commands.

#### Testing specific or multiple server versions
Testing against one of the standard server versions can be done using the `testPostgres<version>` family of tasks. Testing all standard versions of the server is supported using the `testAllPostgresVersions` task which currently tests versions `11`, `10`, `9.6`, `9.5` & `9.4`. The standard version list will be updated as necessary.

Finally testing against a specific set of server versions can be done by executing Gradle with a provided list of versions in `postgresVersions` as in `./gradlew -PpostgresVersions=10,9 test`

#### Disabling Docker
To test against a local or otherwise specific version of the server docker can be disabled by passing Gradle the `noDocker` property (e.g. `./gradlew -PnoDocker=true test`).  It will use the standard test suite properties for server, port, etc.

### OSGI
The entire build is converted with the exception of `OSGI` packaging. The maven build’s OSGI packaging was already incorrect and produced invalid bundles. This will be added back in an upcoming change that tackles it correctly.


jesperpedersen:
I'm quite ok with switching to gradle

jesperpedersen:
Top level `README.md` and `SETTINGS.md` went away

kdubb:
Yup. Noticed just after I merged of course. My asciidoc changes got mixed in unfortunately. They’ve been added back.

PR #371, opened 2019-01-30
Add proper documentation

kdubb:
* Adds `asciidoc` documentation
* `documentation` module added
  * Collects generated docs from all projects (e.g. settings.adoc from driver)
  * Generates HTML “User Guide” & “Release Notes” using `asciidoctor`.
  * Generated HTML docs pushed to `gh-pages`


jesperpedersen:
Nice job !

kdubb:
@jesperpedersen I just updated the gh-pages pages site. I riffed on the simple JUnit site. All of the information about downloads, etc. is now in the user guide. So, this ushers them to it; rather than maintaining a redundant set of information.

For releases I plan on relying on GitHub releases, each release can point to the related user-guide & release-notes as it is archived by version in gh-pages.  It also means a simple tag & pasting of links updates all the "release" documentation.

kdubb:
@jesperpedersen Currently it is using 0.8 as "current" but the documentation push from travis actually continually pushes to "snapshot" for versions with "SNAPSHOT" and then pushes as both "$version" and "current" when it's a release version number.

jesperpedersen:
Sounds good

Issue #376, opened 2019-02-01
Transfer wiki to documentation


kdubb:
All the information in the wiki that is not project related should be moved into the documentation to be versioned and released as normal.

kdubb:
I've deleted most of the wiki information, as it is already covered in the documentation.

@jesperpedersen Are the WildFly wiki pages still relevant and up-to-date? If I'll transfer them to docs and delete the pages.

jesperpedersen:
Yeah, they should be - apart from the version number of the driver

brettwooldridge:
@kdubb @jesperpedersen It was my impression that the github wiki is essentially a separate git repository, and can be versioned independently of the driver.  For example, the git repo for my github's project wiki is at ``git@github.com:brettwooldridge/HikariCP.wiki.git``.  And I manage it separately as a repository.

jesperpedersen:
Frankly, I think keeping it in the documentation is best. Otherwise there are 2 places where people has to contribute to.

kdubb:
@brettwooldridge Agreed but now that we have automatically built documentation I want to consolidate all that information into a single place. The `gh-pages` push is versioned to the release therefore documentation will be available historically for each release. The wiki would need extra care to keep in sync.

PR #378, opened 2019-02-02
Add Application Servers appendix & WildFly config

kdubb:
Completes #376 

@jesperpedersen A quick once over would be good. I attempted to download WildFly and validate the docs but apparently wildfly.org (actually all of jboss.org) seems to be down at the moment.


kdubb:
@jesperpedersen WildFly is back and I got my hands on the latest version. I'v revamped the instructions to match the new UI and added CLI instructions.

If you can't get it started in WildFly, we're in trouble 😂

PR #382, opened 2019-02-04
COPY support

kdubb:
This adds basic PostgreSQL `COPY` support.

Currently there are two methods two invoke `COPY`...

The first method reads from `System.in` or writes to `System.out`, as the commands themselves reference.   Unfortunately this is not thread safe because to redirect `System.in` or `System.out` you need to call `System.setIn` or `System.setOut` accordingly; these have global ramifications.

The alternative, thread safe, method is to use the new `PGConnection.copyIn` and `PGConnection.copyOut` methods which accept InputStream and OutputStream and does not use system globals.

~~These are _basic_ methods that access the copy functionality.  Before this PR is finished an implementation of an "encoding" input stream and "decoding" output stream will be added that handles translating Java objects to/from the correct format for `COPY`.~~ I've scrapped ideas for encoding and decoding streams right now. As I was implementing it, it dawned on my that `TRUNCATE <table>` takes nearly the same path for data out and the new batching provides a performant method for getting Java data in. Encoding and Decoding streams can be implemented on top of the current functionality and even outside the driver, if needed.

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-0.9.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-0.9.adoc
@@ -24,7 +24,7 @@ on GitHub.
 ==== New Features and Improvements
 
 * Batch executions use pipelined requests to increase performance by as much as 70%
-
+* PostgreSQL COPY FROM/TO support
 
 [[release-notes-0.9-pgjdbc-ng-udt]]
 === PGJDBC-NG UDT Generator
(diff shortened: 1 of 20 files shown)


tlf30:
Hello, 
I am wondering on the status of this branch. I am needing this functionality and am wondering when it will be available. 

Thank you,
Trevor Flynn

kdubb:
@tlf30 I'll be merging this to mainline very soon. OpenJDK 11 on ubuntu is presenting issues unrelated to copy support that I am working today to solve; once that's done it'll be merged and available in 0.9-SNAPSHOT

tlf30:
@kdubb, sounds good, thank you!

kdubb:
@tlf30 Support is available in 0.9-SNAPSHOT now

tlf30:
Thank you!

tlf30:
@kdubb I am not seeing the snapshot in maven, has it been uploaded to the maven repo?

kdubb:
Snapshots are deployed to maven central snapshots repository... oss.sonatype.org/content/repositories/snapshots

tlf30:
Thank you,
I forgot I needed to enable snapshots.

PR #394, opened 2019-04-09
Enhance the number to bigdecimal conversion. Previously conversion ex…

tle211212:
…ception is thrown when passing int to a param of preparedstatement while its database type is decimal.

### DIFF
--- a/driver/src/main/java/com/impossibl/postgres/system/procs/Numerics.java
+++ b/driver/src/main/java/com/impossibl/postgres/system/procs/Numerics.java
@@ -177,7 +177,18 @@ public class Numerics extends SimpleProcProvider {
         return;
       }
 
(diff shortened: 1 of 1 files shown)


kdubb:
This is not the correct fix... the `encodeNativeValue` method is only required to handle the procs "native" value; in this case `BigDecimal`.   The `convertInput` method above should have already been called and converted any `Number` to `BigDecimal`.

Can you provide details on how/why that's not happening?

kdubb:
@tle211212 I added tests for this case and the issue was brought to light. I add PR #395 with the proper fix.

Regardless, thanks for your contribution!

kdubb:
@tle211212 #395 is now available in 0.9-SNAPSHOT. If it doesn't work for you please open another PR or issue.

PR #398, opened 2019-04-15
Fix bug with converting LocalTime.MIN

purpleP:
long is unsigned in java, so `MIN_VALUE` is some negative numer. `0L` should be used instead as a minimum value.


purpleP:
@kdubb Is this gonna be merged? What's blocking pull request?
By the way I don't see the need to check for `LocalTime.MIN` and `LocalTime.MAX` as special cases. Just convert `LocalTime` to microseconds.

kdubb:
@purpleP The problem is that `convertInput` is incorrect. It uses `LocalTime.MAX`/`LocalTime.MIN` to signify infinity; essentially this is a copy/paste type error from Local & Offset DateTime handlers where it made sense.

We need an alternative for infinity to be returned and the tests you change should check for those values instead.

PR #431, opened 2019-09-27
Proper handling of client/server locales

kdubb:
## Overview
PostgreSQL’s text output is (mostly) never affected by the server’s locale. Previously the server locale was mapped and the client was using it to parse numeric and monetary values. This doesn’t work since the server always outputs in ‘C’ locale.

Secondarily JDBC has conversion requirements from String to numbers, dates, etc. These must be done in the JVM’s locale, or by an API specified locale when available.

## Solution
`Context` now users client & server specific formatters for integers, decimals, currency, dates, times and timestamps. All of the `Proc` encoders and decoders now use the correct formatting method based on where the string/text is coming from.

Server formatters are “locked” to the ‘C’/‘POSIX’ locale matching the server, usually utilitizing Java’s default parsing/formatting. The client formatters are initialized utilizing the JVM’s current locale during connection.

This has the effect of properly translating JDBC text input in non ‘C’ locales to/from their ‘C’ locale format required for PostgreSQL’s “text” value format.  This happens when utilizing JDBC casting and together with PostgreSQL’s “text” value formats.

## Locale ID/Name Parsing
The locale parsing has been rewritten and should now handle a variety of input types including Win32 verbose names. The Win32 parsing now properly requests language & country in English to ensure it maps properly.

When it cannot find a matching locale it now emits a warning and defautls to the ‘C’/‘POSIX’ locale. This is because locale parsing is no longer required for for normal driver operation (as it should have been all along). It is currently only used for the `money` type (which has problems, see below).

## Money Type
Note that the `money` type is unique and PostgreSQL outputs it according to the `lc_monetary` variable. This changes the numeric formatting symbols as well as adds a currency symbol.  This means that the “server” currency formatter is **not** locked to ‘C’ but is assigned a value matching the server’s `lc_monetary`.

The `money` type seems generally to be “wrong” changing monetary locales causes values to change currency (e.g. from USD to EUR) without any exchange (which would be nearly impossible).  It is assumed you probable shouldn’t be using this type and you certainly shouldn’t be relying on the server’s output to properly format it in a “currency”.

### DIFF
--- a/driver/src/main/java/com/impossibl/postgres/system/BasicContext.java
+++ b/driver/src/main/java/com/impossibl/postgres/system/BasicContext.java
@@ -83,6 +83,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.TimeZone;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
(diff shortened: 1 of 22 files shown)


andreak:
I have to ask; Are you going to merge this and roll a 0.9 release soon-ish? :-)

kdubb:
Yessir.  I am working on some other outstanding issues and will release a 0.8.3 in the coming days. This is an important fix that covers quite a few issues.

andreak:
Great!

Issue #422, opened 2019-07-26
Url parameters not decoded


futpib:
It is impossible to use a password with `&` in it with this driver.

The regexp below does not allow for `<user>:<password>@` syntax (even though documentation does advertise it https://impossibl.github.io/pgjdbc-ng/docs/0.8.2/user-guide/#connection-urls)

https://github.com/impossibl/pgjdbc-ng/blob/239d492d1b244974b7c6356f5fee329839087989/driver/src/main/java/com/impossibl/postgres/jdbc/ConnectionUtil.java#L255-L256

If you pass it like `password=foo&bar`, `bar` will be interpreted as a seconds parameter. If you properly escape it like `password=foo%26bar`, `%26` will never be decoded to `&` and password will never match.

https://github.com/impossibl/pgjdbc-ng/blob/239d492d1b244974b7c6356f5fee329839087989/driver/src/main/java/com/impossibl/postgres/jdbc/ConnectionUtil.java#L311-L325

kdubb:
#433 Fixes this by allowing authority based `user:pass` and decoding escapes in parameter names/values.  In either method percent escaping should be used for special characters.

PR #433, opened 2019-09-27
Connection URL fixes/enhancements

kdubb:
* Allow user/pass as part of URL authority; as is outlined in docs ;)
* Allow user/pass to be complex values using percent encoding
* Allow query parameter names/values to be complex using percent encoding


kdubb:
Fixes #422

Issue #421, opened 2019-07-21
com.impossibl.postgres.system.Version.parse wrongly assumes some specific version format and fails with fat-jar builds


amyodov:
When launching, the `PGDriver` reads the package implementation version: (https://github.com/impossibl/pgjdbc-ng/blob/develop/driver/src/main/java/com/impossibl/postgres/jdbc/PGDriver.java#L78):

~~~java
  static {
    String version = PGDriver.class.getPackage().getImplementationVersion();
    version = version != null ? version : "0.0.0-DEVELOP"; // Ensure it works when in IDE
    VERSION = Version.parse(version);
  }
~~~

According to the `Package#getImplementationVersion` specification, the version string “*does not have any particular syntax specified or expected by the Java runtime*”. Nevertheless, the `Version.parse` call later on (implemented at https://github.com/impossibl/pgjdbc-ng/blob/develop/driver/src/main/java/com/impossibl/postgres/system/Version.java#L48) tries to match it with a very specific format defined by a very specific regex.

When the application is built as a “fat jar”, having just a single JAR for all the code, this version parser actually reads the version of the pgjdbc-ng-using application, rather than the pgjdbc-ng itself. And this version may be defined very randomly (remember “does not have any particular syntax specified”), in my case it is something like `"1.0.8.2019-05-22-1838"`. This causes pgjdbc-ng fail completely when used from this far-jar application.

kdubb:
This is fixed in #434 by generating a dedicated `pgjdbc-ng.properties`. Assuming your fat-jar pulls in all of the resources in `META-INF` (which it better be doing or else it won't work anyway).

PR #434, opened 2019-09-27
Load driver build properties from resource

kdubb:
This _should_ ensure that even if the packaging (e.g. MANFIEST.MF) changes that the driver name/version will always be reported correctly.


kdubb:
Fixes #421

Issue #430, opened 2019-09-17
PGNotificationListener duplicate calls to close()


paol:
When using the notification listeners feature, the closed() callback gets called simultaneosly in 2 different threads when the underlying connection is closed. For some reason this only happens on the second and subsequent connections, not the first connection.

A minimal sample to repro the bug:
```java
package test;

import java.sql.Connection;
import java.sql.DriverManager;
import com.impossibl.postgres.api.jdbc.PGConnection;
import com.impossibl.postgres.api.jdbc.PGNotificationListener;

public class App
{
    public static void log(String msg) {
        System.out.println(String.format("%d [%20s] %s",
                System.currentTimeMillis(), Thread.currentThread().getName(), msg));
    }

    private static void runOnce() throws Exception {
        log("----- running one iteration -----");

        Connection conn = DriverManager.getConnection(
                "jdbc:pgsql://localhost/test-db", "postgres", "postgres");
        PGConnection pgconn = conn.unwrap(PGConnection.class);

        Listener listener = new Listener();
        pgconn.addNotificationListener(listener);
        pgconn.close();
    }

    public static void main(String[] args) throws Exception
    {
        runOnce();
        runOnce();
        runOnce();
        runOnce();
    }
}

class Listener implements PGNotificationListener {
    @Override
    public void notification(int processId, String channelName, String payload) {
        App.log("notification " + processId);
    }
    @Override
    public void closed() {
        App.log("closed");
    }
}
```

This prints the following (notice the double call to close on the second and following iterations):
```
1568722592552 [                main] ----- running one iteration -----
1568722593190 [                main] closed
1568722593190 [                main] ----- running one iteration -----
1568722593216 [                main] closed
1568722593216 [     PG-JDBC I/O (1)] closed
1568722593216 [                main] ----- running one iteration -----
1568722593236 [     PG-JDBC I/O (1)] closed
1568722593236 [                main] closed
1568722593236 [                main] ----- running one iteration -----
1568722593259 [     PG-JDBC I/O (1)] closed
1568722593259 [                main] closed
```

pgjdbc-ng version: 0.8.2
PostgreSQL version: 9.6.15
JDK version: 11

kdubb:
Fixed in #440

PR #440, opened 2019-09-28
Stop multiple notification listener `close` calls

kdubb:
* Reports close only from the I/O thread’s close signal
* Adds test to ensure close is only called once per connection


kdubb:
Fixes #430

Issue #131, opened 2015-02-05
broken inserting of interval from string


Redeeman:
It appears inserting intervals like this:
try( PreparedStatement ps = conn.prepareStatement( "INSERT INTO urls_loads (url_id, loadtime) VALUES(?, ?::interval)" ) ) {
    ps.setInt( 1, utm.id );
    long touse = diff/1000;
    ps.setString( 2, "'" + touse + " microseconds'" );
    ps.executeUpdate();
}

loadtime column is of type interval, and printing out touse using system.out.println gives something like:
'4647685 microseconds'

running query in psql: SELECT  '4647685 microseconds'::interval
works, and the jdbc-code shown above also works with the "original" postgres jdbc driver

Server used is 9.4, and what actually happens is that it just goes in as "00:00:00"

kdubb:
This is a bug in that the `String` to interval conversion is not correct.

To be correct this also requires implementing ISO interval parsing/formatting along with the Postgres parsing/formatting

kdubb:
Fixed in #441, note that clients must be passing string intervals in ISO 8601 format; unless they are relying on server parsing.

PR #441, opened 2019-09-29
Interval overhaul

kdubb:
## Interval class
The `Interval` type has been completely replaced with a verions based on Java 8’s `java.time` package.  The file is based on the fully featured `PeriodDuration` class from [three-ten](https://github.com/ThreeTen/threeten-extra).

## Server/Client parsing
Like Numeric & Date parsing/formatting the “server” parsing is matched to the server based on the `IntervalStyle` server setting.  The client parsing is independently set to ISO.  This means that all JDBC text input for intervals is in ISO 8601 format and doesn’t accept Postgres’s interval format.

To pass data to the server in Postgres format it can be done be passing it as text and casting it to an interval (or relying on implicit casting) on the server.


kdubb:
Fixes #131

Issue #447, opened 2019-10-10
CallableStatement cannot be used as a PreparedStatement


kdubb:
Liquibase uses a `CallableStatement` but never assigns any parameter modes and calls `executeQuery`; thus treating it exactly like a `PreparedStatement`.   This is demonstrated in the test case presented in #413 but surely any use of Liquibase will encounter this issue.

The driver treats it as a `CallableStatement` regardless and throws away any result sets returned by the query. It should detect this situation and revert to its `PreparedStatement` functionality.

I have no idea what the JDBC specification expects to happen in this scenario as it doesn't appear to address it.  So altering the functionality to match the mainline driver and that of a popular SQL library is probably the best coarse.

PR #449, opened 2019-10-10
Allow executeQuery to operate normally on CallableStatement

kdubb:
When calling `executeQuery` on a CallableStatement, if no parameters modes have been assigned, treat it as  a normal `PreparedStatement`.  This makes its operation more compatible with the mainline driver and some libraries expected usage.


kdubb:
closes #447

Issue #454, opened 2019-10-29
SQLException/ConversionException thrown with select datacl from pg_database


kwesterfeld:
If there are database permissions on a database that are granted to a role having the WITH GRANT OPTION, this is not parsable by the pgjdbc-ng driver (although it seems the 0.5 version worked).

The issue is that the acl item looks like this, and the regex in ACLItem.java does not allow for an asterisk:

```
yellowbrick=# select datname,datacl from pg_database;
        datname        |                                 datacl
-----------------------+------------------------------------------------------------------------
 template0             | {=c/kwesterfeld,kwesterfeld=CTc/kwesterfeld}
 _YBPG_stats           |
 template1             | {=c/kwesterfeld,kwesterfeld=CTc/kwesterfeld}
 yellowbrick           | {=Tc/kwesterfeld,kwesterfeld=CTc/kwesterfeld,ybd_test=CTc/kwesterfeld}
 yellowbrick_test      | {=Tc/yellowbrick,yellowbrick=CTc/yellowbrick,ybd_test=CTc/yellowbrick}
 yellowbrick_test_utf8 | {=Tc/yellowbrick,yellowbrick=CTc/yellowbrick,ybd_test=CTc/yellowbrick}
 moo                   | {kwesterfeld=CTc/kwesterfeld,foo=CTc*/kwesterfeld,too=T/kwesterfeld}
(7 rows)

```

NB: the "moo" database has the grant option where asterisk appears in the acl.

Stack trace: 

```
Caused by: java.sql.SQLException: com.impossibl.postgres.system.ConversionException: Unable to convert value to class com.impossibl.postgres.api.data.ACLItem
	at com.impossibl.postgres.jdbc.PGBuffersArray.getArray(PGBuffersArray.java:171)
	at com.impossibl.postgres.jdbc.PGArray.getArray(PGArray.java:91)
	at io.yellowbrick.smc.db.RowSet.processRow(RowSet.java:99)
	at io.yellowbrick.smc.db.impl.Table$1.processRow(Table.java:375)
	at org.springframework.jdbc.core.JdbcTemplate$RowCallbackHandlerResultSetExtractor.extractData(JdbcTemplate.java:1566)
	at org.springframework.jdbc.core.JdbcTemplate$1.doInPreparedStatement(JdbcTemplate.java:693)
	at org.springframework.jdbc.core.JdbcTemplate.execute(JdbcTemplate.java:629)
	... 88 more
[...]

PR #460, opened 2019-11-06
Fix ACLItem parsing and enhance public API

kdubb:
ACLItem now supports the “grant option” on privileges.  The new API terminology follows that of PostgreSQL’s source code; a privilege is still a “privilege” but a privilege and a grant option together are known as a “right”.

#### API
* `getRights`/`setRights` methods added to access full scope of ACL options
* new constructor that takes a `Right[]`
* `rightsOf` and `priviliegesOf` utilities to convert to/from string formats

Backwards compatibility is maintained and getting/setting “privileges” is still allowed along with the new “rights” methods.


kdubb:
fixes #454

Issue #417, opened 2019-06-13
intermittent StackoverflowError on system startup


robertfw:
We are seeing a StackoverFlow error occasionally when starting up a c3pO connection pool. I've been trying to suss out more information but haven't been able to come up with anything repeatable. Any thoughts?

We are giving c3pO a connection-uri of `jdbc:pgsql://our-host/our-db?user=our-user&password=our-password&application_name=our-application.label`. I've tried leaving out the `application_name` as that was a relatively recent addition, but the error occurs both with and without it.

c3pO restarts the thread and typically the system will startup properly on the second try, but not always. occasionally the application needs a full restart before it will come up properly.

The trace we see is

```
Exception in thread "C3P0PooledConnectionPoolManager[identityToken->z8kfsxa3mrl5xe1phwf9l|4838e3ad]-HelperThread-#0" java.lang.StackOverflowError
	at java.lang.StringBuilder.append(StringBuilder.java:136)
	at com.impossibl.postgres.system.Setting.getSystem(Setting.java:710)
	at com.impossibl.postgres.system.AbstractContext.getSetting(AbstractContext.java:35)
	at com.impossibl.postgres.system.BasicContext.getSetting(BasicContext.java:234)
	at com.impossibl.postgres.system.procs.Strings$TxtDecoder.decodeNativeValue(Strings.java:238)
	at com.impossibl.postgres.system.procs.Strings$TxtDecoder.decodeNativeValue(Strings.java:221)
	at com.impossibl.postgres.system.procs.AutoConvertingTextDecoder.decodeValue(BaseDecoders.java:287)
	at com.impossibl.postgres.system.procs.BaseTextDecoder.decode(BaseDecoders.java:218)
	at com.impossibl.postgres.system.procs.Strings$TxtDecoder.decode(Strings.java:221)
	at com.impossibl.postgres.system.procs.BaseTextDecoder.decode(BaseDecoders.java:196)
	at com.impossibl.postgres.system.procs.Strings.convertOutput(Strings.java:129)
	at com.impossibl.postgres.system.procs.Strings.access$100(Strings.java:49)
	at com.impossibl.postgres.system.procs.Strings$TxtDecoder.lambda$new$0(Strings.java:224)
	at com.impossibl.postgres.system.procs.AutoConvertingTextDecoder.convertOutput(BaseDecoders.java:247)
[...]

PR #461, opened 2019-11-06
Stop sharing of Proc instances

kdubb:
For diagnosing #417, this disables the explicit `Proc` instance sharing.

With this change, disabling registry sharing _should_ disable all sharing of anything.

Long term, this really shouldn’t have an impact on startup time. If registries are being shared they should only endup with a single `Proc` instance anyway.


kdubb:
Replaced with #509

PR #476, opened 2020-02-07
Gradle 6 upgrade

moio:
Hi from the openSUSE project Uyuni :wave:!

As (Linux) distros, we get to package all software we ship, in our case, in the RPM format. We have a need to have reproducible offline builds of our packages, and Gradle has a history of making our life a bit difficult in that respect.

Fortunately though, the very latest versions of Gradle now feature relocatable caches, meaning you can build a project from a machine with Internet access, zip its cache and then later repeat the build from a different host even if completely isolated from any network.

So here I am, proposing to upgrade Gradle to version 6 with this PR :sweat_smile: 

I admit I am a total newbie when it comes to Gradle and Kotlin, so things might have escaped my attention, but so far at least build and tests work from my machine. I'd appreciate if you could take the time to have a look and see if this is merge-able as it is or what needs to be done in order to be on the latest Gradle.

Thank you in advance!

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,5 +1,5 @@
 
-import com.github.breadmoirai.GithubReleaseTask
+import com.github.breadmoirai.githubreleaseplugin.GithubReleaseTask
 import de.undercouch.gradle.tasks.download.Download
 
 plugins {
@@ -63,7 +63,6 @@ tasks {
 
   named<GithubReleaseTask>("githubRelease") {
     dependsOn(downloadArtifacts)
-    setToken(project.properties["github.token"]?.toString() ?: "")
     setOwner("impossibl")
     setRepo("pgjdbc-ng")
     setTagName("v$version")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -63,6 +63,7 @@ tasks {
 
   named<GithubReleaseTask>("githubRelease") {
     dependsOn(downloadArtifacts)
+    setAuthorization("token ${project.properties["github.token"]?.toString() ?: ""}")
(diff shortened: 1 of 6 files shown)


build.gradle.kts
```diff
@@ -63,7 +63,6 @@ tasks {
 
   named<GithubReleaseTask>("githubRelease") {
     dependsOn(downloadArtifacts)
-    setToken(project.properties["github.token"]?.toString() ?: "")
```
kdubb, line 66: Is there a reason you removed the setting of the token? I'm assuming this shouldn't affect your builds as you shouldn't be attempting a GitHub release.

buildSrc/build.gradle.kts
```diff
@@ -1,10 +1,9 @@
 
 plugins {
   id("java-library")
-  id("org.jetbrains.kotlin.jvm") version "1.3.11"
+  id("org.jetbrains.kotlin.jvm") version "1.3.30"
```
kdubb: Current plugin is at `1.3.61`. Can we just move up to that?

kdubb:
Thanks for the PR! 👏 

Just a couple simple questions but it looks correct in general.

build.gradle.kts
moio: Sure, IIUC the method was removed from the plugin, which will fetch the `github.token` property by default, so the net effect should be no change on your end.

Of course I could not really test this.

buildSrc/build.gradle.kts
moio: I have no objection, just tried to break as little as possible. Feel free to bump further!

PR #477, opened 2020-02-10
Fixes #465.   Corrections for quoted identifiers and embedded double quotes

### LINKED ISSUE #465: SQL containing identifier that begins and ends with double quotes fails with "unterminated string literal"
An identifier with double quotes issue has been raised with our database
which uses the pgjdbc-ng driver.  A table can be created that contains double
quotes at the beginning and and of the identifier, and pgdjbc-ng either throws
an ArrayIndexOutOfBoundsException, or throws "Unterminated string literal",
depending on the SQL also adding double quotes around the identifier or not.

I have modified the unit test in the repo to resolve: com.impossibl.postgres.jdbc.PreparedStatementTest.testDoubleQuotes

Adding the patch below for the test.  I could not see an easy fix in the double
quote parser or I would send a PR.

The test results in a failure this way:

```
  Test testDoubleQuotes FAILED

  java.sql.SQLException: Error parsing SQL at position 28 (CREATE TABLE """with_quotes""" (i integer)): Unterminated string literal
      at com.impossibl.postgres.jdbc.PreparedStatementTest.testDoubleQuotes(PreparedStatementTest.java:356)
```

Patch for test:

```
diff --git a/driver/src/test/java/com/impossibl/postgres/jdbc/PreparedStatementTest.java b/driver/src/test/java/com/impossibl/postgres/jdbc/PreparedStatementTest.java
index 891f6081..d32b5ada 100644
--- a/driver/src/test/java/com/impossibl/postgres/jdbc/PreparedStatementTest.java
+++ b/driver/src/test/java/com/impossibl/postgres/jdbc/PreparedStatementTest.java
@@ -350,7 +350,7 @@ public class PreparedStatementTest {
 
   @Test
   public void testDoubleQuotes() throws SQLException {
-    String[] testStrings = new String[] {"bare ? question mark", "single ' quote", "doubled '' single quote", "doubled \"\" double quote", "no backslash interpretation here: \\", };
+    String[] testStrings = new String[] {"bare ? question mark", "single ' quote", "doubled '' single quote", "doubled \"\" double quote", "no backslash interpretation here: \\", "\"\"with_quotes\"\"", " \"\"with_quotes\"\" ", " \"\"with_quotes\"\"", "\"\"with_quotes\"\" ", };
 
     for (int i = 0; i < testStrings.length; ++i) {
       PreparedStatement pstmt = conn.prepareStatement("CREATE TABLE \"" + testStrings[i] + "\" (i integer)");
```

kwesterfeld2:
(1) add a bunch of tests for double quote insanity: trailing, leading, and
    intermixed double quotes in identifiers.
(2) alter SQLText::consumeQuotedIdentifier() to understand not skipping
    quoted identifiers unless adjacent quotes encountered (ie. not processing
    an odd number in a row), and to do look-ahead for adjacent double quotes that
    do not terminate an identifier.

### DIFF
--- a/driver/src/main/java/com/impossibl/postgres/jdbc/SQLText.java
+++ b/driver/src/main/java/com/impossibl/postgres/jdbc/SQLText.java
@@ -365,16 +365,22 @@ public class SQLText {
   }
 
   private static int consumeQuotedIdentifier(final String sql, final int start, final CompositeNode parent) throws ParseException {
-    int ndx = start + 1;
+    int ndx = start + 1, adjacentQuoteCount = 0;
     do {
       char c = sql.charAt(ndx);
       if (c == '"') {
-        if (sql.charAt(ndx - 1) == '"') {  // look-behind
+        ++adjacentQuoteCount;
+        if (sql.charAt(ndx - 1) == '"' && (adjacentQuoteCount % 2) == 0) {  // look-behind
           // skip escaped
         }
         else {
-          break;
+          if ((ndx == sql.length() - 1 || sql.charAt(ndx + 1) != '"')) {  // look-ahead
+            break;
+          }
         }
+      }
+      else {
+        adjacentQuoteCount = 0;
       }
 
       if (lookAhead(sql, ndx) == 0) {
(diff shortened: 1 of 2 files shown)


kdubb:
@kwesterfeld2 Thanks for the PR!

PR #515, opened 2020-11-17
Don't query for SSL support if SSLMode >= required

### LINKED ISSUE #514: Unable to force SSL connections
Postgres has a feature that forces clients to use SSL connections.

`pgjdbc-ng` doesn't work with that feature due to the query for SSL before creating the SSL factory.

See here: https://github.com/impossibl/pgjdbc-ng/blob/b01be976d422dcb48e69d0243f9e6cf5f1d0dc63/driver/src/main/java/com/impossibl/postgres/protocol/v30/ServerConnectionFactory.java#L155-L164

I think that either this check should not be executed when SSLMode is required OR the error code for requiring SSL should be caught and recognized here. Is there any preference for which path should be taken?

rockwotj:
Since SSL is required, there is not really a ton of benefit to query if SSL is supported. 
Let me know if there are better ways of doing this! Thanks.

Fixes #514

### DIFF
--- a/driver/src/main/java/com/impossibl/postgres/protocol/v30/ServerConnectionFactory.java
+++ b/driver/src/main/java/com/impossibl/postgres/protocol/v30/ServerConnectionFactory.java
@@ -154,44 +154,39 @@ public class ServerConnectionFactory implements com.impossibl.postgres.protocol.
 
       if (sslMode != SSLMode.Disable && sslMode != SSLMode.Allow) {
 
-        // Execute SSL query command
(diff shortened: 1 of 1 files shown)


kdubb:
While it looks like a plausible change it suggests that the sever does _not_ require an SSL query when the mode is `require`. Are we sure that's correct?  I thought the server still required the query.

kdubb:
Should've looked at CI first... the SSL tests are failing in certain modes.

rockwotj:
Sorry I think I was mistaken about what was going on here, I need to do more debugging. Thank you for your time.

PR #524, opened 2020-12-28
Fix #523: Use the appropriate type map in Struct.getAttributes

### LINKED ISSUE #523: Struct#getAttributes seems to ignore any provided type map
When the overload of `getAttributes` that takes a type map on a `Struct` is called, the mappings do not appear to have any effect. For example:
```java
    try (PreparedStatement st = conn.prepareStatement("SELECT ROW(now())")) {
      try (ResultSet rs = st.executeQuery()) {
        assertTrue(rs.next());
        Object[] resultWithTypeMap = rs.getObject(1, Struct.class).getAttributes(Collections.singletonMap("timestamptz", OffsetDateTime.class));
        Object[] resultWithoutTypeMap = rs.getObject(1, Struct.class).getAttributes();

        System.out.println(resultWithTypeMap[0].getClass());
        System.out.println(resultWithoutTypeMap[0].getClass());
      }
    }
```
produces:
```
class java.sql.Timestamp
class java.sql.Timestamp
```

This _does_ seem to work as expected when dealing with an array, i.e.:
```java
    try (PreparedStatement st = conn.prepareStatement("SELECT ARRAY [now()]")) {
      try (ResultSet rs = st.executeQuery()) {
        assertTrue(rs.next());
        Object[] resultWithTypeMap = (Object[]) rs.getArray(1).getArray(Collections.singletonMap("timestamptz", OffsetDateTime.class));
        Object[] resultWithoutTypeMap = (Object[]) rs.getArray(1).getArray();

        System.out.println(resultWithTypeMap[0].getClass());
        System.out.println(resultWithoutTypeMap[0].getClass());
      }
    }
```
produces:
```
class java.time.OffsetDateTime
class java.sql.Timestamp
```

Looking at the code, it seems there is some special handling in the `PGBuffersArray` which calls `lookupCustomType`. However, in the `PGBuffersStruct` case, the code path doesn't seem to consider the type map, and we fall back to the default type of the decoder.

I'd be happy to try and create a PR for a fix, but wanted to first make sure I'm not using this wrong or doing something unreasonable!

tinselspoon:
Any custom type map registered with the connection or `getAttributes` call is now used.

Also adds a short note to the documentation on the behaviour of the method when using `Statement`/`PreparedStatement`.

### DIFF
--- a/documentation/src/docs/asciidoc/user-guide/jdbc.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/jdbc.adoc
@@ -63,6 +63,10 @@ the standard `java.sql.Struct` API or via custom mapped Java classes that implem
 {drivername} fully supports UDTs via the JDBC `java.sql.Struct` API for specifying UDT parameters
 and for retrieving UDT values in from ``ResultSet``s.
 
+NOTE: Strong typing on the elements returned from `Struct.getAttributes()` will only be present if the
+query was run from a `PreparedStatement`. A regular `Statement` uses the text-based {vendorname} protocol,
+which will always return the values of the UDT represented as strings.
+
 ==== java.sql.SQLData
 
 Alternatively to using the standard Struct interface to access UDTs, custom Java classes can be created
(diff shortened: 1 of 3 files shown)


kdubb:
@tinselspoon Can you merge these into a single commit and update the branch?

tinselspoon:
@kdubb All done :)

kdubb:
@tinselspoon 👍 Thanks for helping out!

PR #549, opened 2021-06-03
Fix Local support for Norwegian Bokmal

defaste:
Connection to a PostgreSQL database on Windows 7 or newer configured to use Norwegian Bokmal language always fails.
This is a quick-and-dirty solution I should use to make the driver to work.
I'm available to test the official fix when it will be available.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -16,7 +16,7 @@ allprojects {
   }
 
   group = "com.impossibl.pgjdbc-ng"
-  version = "0.8.9"
+  version = "0.8.9-bokmal"
 
   extra["isSnapshot"] = version.toString().endsWith("SNAPSHOT")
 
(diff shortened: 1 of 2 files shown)


kdubb:
@defaste What is the full `localeValue` string passed to `Locales.parseLocale`?

defaste:
> @defaste What is the full `localeValue` string passed to `Locales.parseLocale`?

The full localeValue is "Norwegian Bokmål". The character å is not parsed correctly, i see a square in the log file.
This is a well known issue for Norwegian language, see https://wiki.postgresql.org/wiki/Changes_To_Norwegian_Locale

kdubb:
> What was changed?
> 
> To fix these issues, it was decided to map the problematic "Norwegian (Bokmål)_Norway" locale name to an alias of the same locale that Windows also recognizes, "Norwegian_Norway".
>
> Unfortunately, even though the code was changed to use the "Norwegian_Norway" alias, the old problematic name will still exist in the pg_database system catalog in old installations, and also in the postgresql.conf configuration file. These instances need to be converted manually.

Shouldn't the database be converted as it says?  Also, if we were to add this very special case, shouldn't we map it to `Norwegian_Norway`?

defaste:
Unfortunately this is a clean recent installation of PostgreSQL 13.3 on Windows 10, official EDB installer.
I can confirm setting localeValue = "Norwegian_Norway" works perfectly in my application.

kdubb:
@defaste Can you remove the version change... I cannot merge until the local remap is the only change.

Issue #562, opened 2021-09-16
Second and subsequent calls to free() on PGBuffersArray throw NPE rather than being no-op


JTEgan:
The PGBuffersArray class implements java.sql.Array, and the doc for the free() method in that interface states:

> If <code>free</code> is called multiple times, the subsequent calls to <code>free</code> are treated as a no-op. 

In the pgjdbc-ng implementation, the second call results in a NullPointerException:

> java.lang.NullPointerException: null
>         at com.impossibl.postgres.jdbc.PGBuffersArray.free(PGBuffersArray.java:294)
>         at...

I have a small change to address the issue, will submit a PR shortly.

PR #563, opened 2021-09-16
Fixes issue #562, multiple calls to PGBuffersArray.free() produces NullPointerException

JTEgan:
API doc for java.sql.Array says that when multiple calls are made to free(), successive calls should be a no-op. This change makes the PGBuffersArray class adhere to that spec.

### DIFF
--- a/driver/src/main/java/com/impossibl/postgres/jdbc/PGBuffersArray.java
+++ b/driver/src/main/java/com/impossibl/postgres/jdbc/PGBuffersArray.java
@@ -291,8 +291,10 @@ public class PGBuffersArray extends PGArray {
 
   @Override
   public void free() {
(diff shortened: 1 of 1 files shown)


kdubb:
@JTEgan Thanks for the contribution.  Can you add a test cast in `ArrayTest` to verify this?

JTEgan:
Yep, will do

kdubb:
@JTEgan Please push a new (or empty) commit to restart actions.

JTEgan:
yeah geez it's been 2 years, I gotta get around to writing that test ;-)

What does it mean to `restart actions`?

kdubb:
The action storage has expired. So I don't know the outcome of them. A new commit (even if just empty) will re-run them.

PR #571, opened 2022-02-25
Add missing quotes around column names

safris:
This portion of the code is executed when PGResultSet.updateRow() is called. The omission of double-quotes around the column names seems to be a typo.


kdubb:
@safris Please push a new (or empty) commit to restart actions.

Issue #579, opened 2022-08-12
What's the status and direction of the project?


raphw:
Hello,
thanks for the work on this driver and leveraging asynchronous notify/listen. Unfortunately, the library relies on several outdated dependencies. Is it that nobody maintains this code anymore?
Thanks.

kdubb:
Not abandoned but we have migrated all of our usage away from JDBC and therefore my main maintenance motivation has gone away.

I will try to get things updated in the coming week.

raphw:
Thanks for the answer. What are you using now? Vertx?

xzilla:
Perhaps you would consider re-licensing this project under the Postgres license, which would allow others to more easily continue the (very nice) work you have started?

kdubb:
@xzilla I'm definitely open. It's currently the 3 clause BSD so I'm not sure how that's limiting anything. That being said, I've wanted to move it of this defunct organization for a while now.

kdubb:
@raphw We've moved entirely to Quarkus. So yes we're using the reactive PG driver from  the Vert.x platform (which Quarkus is built upon).

davecramer:
@kdubb I have no issue moving it into org.postgresql.<something> and housing it inside our github repo if that's something that appeals to you.

kdubb:
@davecramer That would be a great move. I'll chat with you in the next week or so to get things going. I've got a lot going currently.

kdubb:
FYI, I'm happy to continue maintaining the project as much as I can. It has involved a large amount of effort to start and continue the project from myself and others. I don't want to walk away but as I said I've got lots of time constraints.

Also, the pluggable type engine this driver is built on to support advanced, composite, and custom types might live on; either as an add-on or integrated with the Vert.x driver. Currently the Vert.x driver has little to no support for them.

jesperpedersen:
@davecramer We have talked about this in the past - at least in person :smiley: The question is who will maintain the code in the new repository if @kdubb doesn't have time for it. I'm only working with C and Rust now, so I don't have time to do bigger contributions.

I'm happy to sign over my contributions to a new license if necessary.

davecramer:
@jesperpedersen well that's an excellent question. One possibility is that inside the PostgreSQL org it might get more love. I don't have an answer.

kdubb:
Regardless of maintenance capacity, moving the copyright to a proper organization and possibly relicensing is something that should be done; the obvious first choice would be the PostgreSQL org.

[...]

PR #586, opened 2023-04-03
Support reading SSL config in diskless environments

rockwotj:
We run in k8s with readonly disks and would like to load or secrets like SSL config for the driver from a secret manager on startup instead of assuming they are on disk. This PR allows for the user to override how SSL certs are loaded.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,7 +44,7 @@ jobs:
           job-id: jdk${{ matrix.jdk }}-build-test
           properties: |
             postgresVersions=${{ matrix.pg }}
-          arguments: test -x checkstyleMain -x checkstyleTest -x documentation
+          arguments: test -x checkstyleMain -x checkstyleTest
 
   deploy:
     needs: [check, build]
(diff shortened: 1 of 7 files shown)


davecramer:
Pretty sure this project has been abandoned

kdubb:
I am trying to find time to update this but with all the other tasks on my plate it's hard to find the time.

rockwotj:
Thanks @davecramer - I did read through https://github.com/impossibl/pgjdbc-ng/issues/579

We're going to maintain our own fork for now, but figured I'd at least put this forward in case that ever changes.

kdubb:
@rockwotj I'm more than willing to add maintainers! I'd like to have some review control during the initial term but that would be a big help!

rockwotj:
@kdubb let me see if my company is willing to pitch in! I know the project doesn't build without some hacks right now - so that would need to be fixed. I'm not a gradle expert but that's probably needing resolution first.

kdubb:
The open question from #579 was who would do the maintenance. If we can find one or more to pitch in, I can fix and upgrade Gradle and help others get up to speed. Moving to @davecramer's world would also make more sense then.

kdubb:
Added related comments to discussion in #579

davecramer:
I'm willing to move this under the https://github.com/pgjdbc, but like @kdubb I don't have time to maintain it. So it would have to have maintainers.

rockwotj:
My company decided to not take up the cup of maintenance unfortunately

PR #603, opened 2024-06-07
Update versions and add instructions on publishing to local maven repo

crowmagnumb:
I was being held up from updating my website by the old version so I compiled it myself. This bumps some of the dependency versions and explains how to publish locally.

### DIFF
--- a/buildSrc/src/main/kotlin/Versions.kt
+++ b/buildSrc/src/main/kotlin/Versions.kt
@@ -6,19 +6,19 @@ object Versions {
    * Source
    */
   val javaTarget = JavaVersion.VERSION_1_8
-  const val kotlin = "1.9.10"
+  const val kotlin = "2.0.0"
 
   /**
    * Required Libraries
    */
-  const val netty = "4.1.100.Final"
+  const val netty = "4.1.109.Final"
 
   /**
    * Tooling Libraries
    */
   const val javaPoet = "1.13.0"
   const val argParser = "2.0.7"
-  const val asciidoctorJ = "2.5.10"
+  const val asciidoctorJ = "2.5.13"
 
   /**
    * Testing dependencies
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -6,19 +6,19 @@ object Versions {
    * Source
    */
   val javaTarget = JavaVersion.VERSION_1_8
-  const val kotlin = "1.9.10"
+  const val kotlin = "2.0.0"
```
davidwheeler123: Kotlin 2.0.0 is very new, sometimes it's good to wait a few minor/patch versions before upgrading. But if we do want to upgrade, [it looks like](https://kotlinlang.org/docs/whatsnew20.html#current-k2-compiler-limitations) we should also update to a more recent Gradle (8.3+) as well

davidwheeler123 (commented):
Thanks for the contribution! 
I've added one comment about upgrading to Kotlin 2 - is that something that's important to you?
I'll leave @kdubb to approve

buildSrc/src/main/kotlin/Versions.kt
davidwheeler123: Oops looks like I was looking at an old branch, we're already using gradle 8.4 🤦

kdubb: Yeah, unless there's a reason to move to 2.0. Probably stay on 1.9.x.

Upgrading to 1.9.20 seems fine.

crowmagnumb: No not important at all. I just always try to stay latest on everything while avoiding bleeding edge. I don't even use Kotlin actually, I've been meaning to start switching at least some of my java work to it but haven't gone there yet. So 1.9 is fine with me. :)

crowmagnumb: I changed it to 1.9.20

davidwheeler123 (approved):

